PR #5, opened 2019-04-22
Enable use of VaFacilities in place of AccessToCare

jjennings-va:
Added in VaFacilites
Removed AccessToCare

Ensured E&E integration was not affected by change
Updated tests to use VaFacilities in place of AccessToCare


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/RestFacilitiesClient.java
```diff
@@ -0,0 +1,61 @@
...
+  private final String vaFacilitiesApiKey;
+
+  private final String baseUrl;
+
+  private final RestTemplate restTemplate;
+
+  /** Autowired constructor. */
+  public RestFacilitiesClient(
+      @Value("${va-facilities.api-key}") String vaFacilitiesApiKey,
+      @Value("${facilities.endpoint.url}") String baseUrl,
```
jblefkove-va: I don't think we need `endpoint` in this property... and let's make it `va-facilities.url` to match the above property.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/VaFacilitiesResponse.java
```diff
@@ -0,0 +1,140 @@
...
+    return data;
+  }
+
+  @Data
+  @Builder
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class VaFacility {
```
jblefkove-va: Now that all of these classes are consolidated, we only need the `Va` prefix on the top-level response class. So let's make this `Facility` and the below class `Attributes`.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/VaFacilitiesResponse.java
```diff
@@ -0,0 +1,140 @@
...
+    @JsonProperty("wait_times")
+    private WaitTimes waitTimes;
+
+    @Data
+    @Builder
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @NoArgsConstructor(access = AccessLevel.PRIVATE)
+    @AllArgsConstructor(access = AccessLevel.PRIVATE)
+    @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+    public static final class Address {
```
jblefkove-va: Move these classes out of `VaFacilityAttributes` so we only have one level of inner classes inside `VaFacilitiesResponse`.

community-care-eligibility/src/main/resources/application.properties
```diff
@@ -34,6 +34,7 @@ bing-maps.api-key=unset
 va-facilities.api-key=unset
 community-care.max-drive-time=60
 community-care.max-wait=20
+facilities.endpoint.url=unset
```
jblefkove-va: Let's add the facilities API url to the make-configs script, like we do with the EE url.

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityTest.java
```diff
@@ -63,69 +62,42 @@ public void empty() {
...
+        .thenReturn(Coordinates.builder().latitude(200.00).longitude(100.00).build());
     when(bingMaps.routes(
-            eq(
-                Address.builder()
-                    .street("66 Main St")
-                    .city("Melbourne")
-                    .state("fl")
-                    .zip("12345")
-                    .build()),
+            eq(Coordinates.builder().longitude(100.00).latitude(200.00).build()),
```
jblefkove-va: I think this test would be a little easier to follow if these patient coordinates were defined once at the beginning of the test.

PR #6, opened 2019-04-29
Task/api 1085 compute overall eligibility 2

jjennings-va:
https://vasdvp.atlassian.net/browse/API-1085

Finished logic for determining eligibility. 
Cleaned up community care output
Now only displays facilities within the state


jblefkove-va:
I think we're missing a requirements. I will help you work through this list when I get in this morning:

* patient ICN should be argument to the API

* Let's add a unit test for the urgent care case, since we don't have a good way to test it with real data.

* To facilitate debugging and ease of use, the returned response should include all of the input parameters (patient ICN, address, service type, established patient) along with the patient address as a lat/long.

* Based on https://www.va.gov/opa/pressrel/pressrelease.cfm?id=5187, we need to support two configuration parameters for drive time, and two for wait time. (Currently we have one each.) The link indicates 30 mins drive time for primary/mental-health care, and 60 mins for specialty, and 20 day wait for primary/mental-health care, and 28 days for specialty.

* In my email to Dave last week, I mentioned that we have work around a bug in how facilities API handles mental health care. For mental health services, the health service type is `MentalHealthCare`, but the wait times are listed under `MentalHealth`. So if `MentalHealthCare` is requested, we need to make sure we filter the resulting facilities under `MentalHealth`.

* We should use the effective date on the eligibility info from E&E. That is, we should only use the codes with effective date prior to current time. (In practice, I expect they will always be in the past, but this is something to check and unit test.)

* I think there is an error in the way the separate wait and drive time checks are implemented. If you have one facility that is within the wait limit, but not the drive limit, and another facility within the drive limit, but not the wait limit, you are eligible for community care, but both checks will fail. Based on the terminology used in  https://www.va.gov/opa/pressrel/pressrelease.cfm?id=5187, the reason in the description, if you have any facilities that meet both the wait and drive time criteria, should be "Access Standards". Let's add a unit test for this case.

* If the overall eligibility is false due to failing the wait/drive time check, we should still add something to the description in that case. The reason can probably still be "Access Standards" whether you are eligible or not.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -80,29 +84,32 @@ private static boolean hasServiceType(
...
-    if (Arrays.asList("AK", "AZ", "IA", "NM", "MN", "ND", "OK", "SD", "UT")
-        .stream()
-        .anyMatch(patientAddress.state()::equalsIgnoreCase)) {
-      // No VAMC locations in these states, automatically eligible
-      return true;
-    }
-    // Filter facilities in same state, within a certain drive time and wait time
-    List<Facility> filtered =
+  private boolean computeEligibilityBasedOnDriveTime(List<Facility> facilities, String serviceType) {
+      int driveTime = ((StringUtils.equalsIgnoreCase(serviceType, "primarycare")) || (StringUtils.equalsIgnoreCase(serviceType, "primarycare")) ? maxDriveTime : maxDriveTimeSpecialty);
```
dbrown-va: It looks like the second half of the `||` needs to be mental health care?  Same with the `computeEligibilityBasedOnWaitTime` function below.

While there, would it be worth making 2 configurable enums, `standardCare` and `specialtyCare` and then checking if the serviceType is present in one or the other?  If future care types are added, then only a single place needs to be updated.

jblefkove-va: I think ultimately we will want an enum that maps all of the different service type values that Facilities API supports.

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -24,26 +23,29 @@
 //      example = "SWAGGER_EXAMPLE_ALLERGY_INTOLERANCE"
 //    )
 public final class CommunityCareEligibilityResponse {
-  Boolean communityCareEligible;
+  CommunityCareEligibility communityCareEligibility;
 
-  List<CommunityCareEligibilities> communityCareEligibilities;
+  PatientRequest patientRequest;
```
jblefkove-va: Put the request at the top.

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -24,26 +23,29 @@
...
-  }
+  @Data
+  @Builder
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class PatientRequest {
+    String serviceType;
+
+    String paitentIcn;
```
jblefkove-va: typo

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -24,26 +23,29 @@
...
-    }
-    return communityCareEligibilities;
-  }
+  @Data
+  @Builder
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class PatientRequest {
+    String serviceType;
```
jblefkove-va: Let's make the order here roughly match the order of the search parameters in the controller. (ICN, address and coordinates, serviceType, established)

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -24,26 +23,29 @@
...
-      facilities = new ArrayList<>();
-    }
-    return facilities;
+    Address patientAddress;
+
+    Coordinates patientCoordinates;
+
+    Boolean establishedPatient;
+
+    String description;
```
jblefkove-va: This looks like a copy-paste error.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
-        bingMaps.coordinates(
-            Address.builder()
-                .street(street.trim())
-                .city(city.trim())
-                .state(state.trim())
-                .zip(zip.trim())
-                .build());
-    GetEESummaryResponse response = eeClient.requestEligibility("1008679665V880686");
+      @NotBlank @RequestParam(value = "serviceType") String serviceType,
+      @NotBlank @RequestParam(value = "patientICN") String patientIcn,
```
jblefkove-va: Let's camel-case that `PatientIcn`. Also, this feels like the most important parameter, let's put it at the front of the list.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/BingResponse.java
```diff
@@ -60,7 +60,7 @@
 
     Integer travelDuration;
 
-    Integer travelDurationTraffic;
+    // Integer travelDurationTraffic;
```
jblefkove-va: Let's remove this.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -47,11 +49,17 @@
...
+  private int maxDriveTimeSpecialty;
+
+  private int maxWaitSpecialty;
+
   /** Autowired constructor. */
   @Builder
   public CommunityCareEligibilityV1ApiController(
       @Value("${community-care.max-drive-time}") int maxDriveTime,
       @Value("${community-care.max-wait}") int maxWait,
+      @Value("${community-care.max-drive-time-specialty}") int maxDriveTimeSpecialty,
```
jblefkove-va: Let's make the other parameter `max-drive-time-primary` to make the distinction between the two clearer. (Actually, while we're renaming these, let's also add units, i.e. `max-drive-time-mins-primary` and `max-wait-days-primary`.)

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -47,11 +49,17 @@
 
   private int maxWait;
 
+  private int maxDriveTimeSpecialty;
+
+  private int maxWaitSpecialty;
```
jblefkove-va: Let's make the order of these fields, the order of the constructor parameters below, and the order of the field assignments in the constructor all match. (Also, I don't see these two fields being set in the constructor.)

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
+            .zip(zip.trim())
+            .build();
+    Coordinates patientCoordinates = bingMaps.coordinates(patientAddress);
+    GetEESummaryResponse response = eeClient.requestEligibility(patientIcn);
     List<VceEligibilityInfo> vceEligibilityCollection =
-        response.getSummary() == null
+        response == null || response.getSummary() == null
             ? Collections.emptyList()
             : response
                 .getSummary()
```
jblefkove-va: Let's move this to a separate method and add null checks for the (unlikely) cases of any of these intermediate objects being null.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
+    final String serviceRequestType =
+        equalsIgnoreCase(serviceType, "MentalHealthCare") ? "MentalHealth" : serviceType;
+    Address patientAddress =
+        Address.builder()
+            .street(street.trim())
+            .city(city.trim())
+            .state(state.trim())
+            .zip(zip.trim())
+            .build();
+    Coordinates patientCoordinates = bingMaps.coordinates(patientAddress);
```
jblefkove-va: Some of these variables can be declared later, closer to where they are used. For instance, `patientAddress` and `patientCoordinates` aren't used until after the EE stuff is done.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
+    List<String> eligibilityDescriptions = new ArrayList<String>();
+    List<String> eligibilityCodes = new ArrayList<String>();
+    GregorianCalendar todaysDate = new GregorianCalendar();
+    for (int i = 0; i < vceEligibilityCollection.size(); i++) {
+      int dateComparision =
+          vceEligibilityCollection
+              .get(i)
+              .getVceEffectiveDate()
+              .compare(DatatypeFactory.newInstance().newXMLGregorianCalendar(todaysDate));
+      if (dateComparision == 0 || dateComparision == -1) {
```
jblefkove-va: You can do this more cleanly with the Java 8 date/time API:
```
Instant now = Instant.now();
for (...) {
  if (vceEligibilityCollection.get(i).getVceEffectiveDate().toGregorianCalendar().toInstant().isBefore(now)) {
    ...
  }
}
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
-            .map(
-                vceEligibilityInfo ->
-                    EligibilityAndEnrollmentTransformer.builder()
-                        .eligibilityInfo(vceEligibilityInfo)
-                        .build()
-                        .toCommunityCareEligibilities())
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+    List<String> eligibilityDescriptions = new ArrayList<String>();
+    List<String> eligibilityCodes = new ArrayList<String>();
```
jblefkove-va: The type parameter on the right is redundant.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
+                .filter(vaFacility -> hasServiceType(vaFacility, serviceRequestType))
+                .filter(
+                    vaFacility ->
+                        equalsIgnoreCase(
+                            vaFacility.attributes().address().physical().state().trim(),
+                            patientAddress.state()))
                 .collect(Collectors.toList());
-    log.info("va facilities filtered by service type {}: {}", serviceType, vaFacilitiesResponse);
+    log.info(
+        "va facilities filtered by service type {}: {}", serviceRequestType, vaFacilitiesResponse);
```
jblefkove-va: Let's note in the logged message that we also filtered by state. Also, we're logging `vaFacilitiesResponse` here, which doesn't have any filtering applied... that seems like a mistake. Let's log just the ids from the filtered list:

```
filteredByServiceTypeAndState
     .stream()
     .map(facility -> facility.id())
     .collect(Collectors.toList())
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
             .map(
                 vaFacility ->
                     FacilityTransformer.builder()
-                        .serviceType(serviceType)
+                        .serviceType(serviceRequestType)
                         .build()
                         .toFacility(vaFacility))
             .collect(Collectors.toList());
     facilities.parallelStream().forEach(facility -> setDriveMinutes(patientCoordinates, facility));
+    Boolean communityCareEligible = false;
```
jblefkove-va: Let's extract the calculation of this value to its own method.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -114,66 +120,119 @@ public CommunityCareEligibilityResponse search(
...
+      communityCareEligible = true;
+    } else {
+      eligibilityDescriptions.add("Access-Standards");
+      if (eligibleByAccessStandards(establishedPatient, facilities, serviceRequestType)) {
+        communityCareEligible = true;
+      }
+    }
+    String communityCareDescriptions = String.join(", ", eligibilityDescriptions);
+    CommunityCareEligibilityResponse.CommunityCareEligibility communityCareEligibility =
+        EligibilityAndEnrollmentTransformer.builder()
```
jblefkove-va: `EligibilityAndEnrollmentTransformer` class seems like overkill here, since it accepts a `CommunityCareEligibility` object as input and then returns one. I think all you need here is a method in this class that takes `communityCareEligible` and `communityCareDescriptions` as arguments, and returns a `CommunityCareEligibility`. And then put the checks that the transformer is doing in there.

community-care-eligibility/src/main/resources/application.properties
```diff
@@ -32,8 +32,11 @@ spring.jackson.default-property-inclusion=non_empty
 #included-references.practitioner=true
```
dbrown-va: There is a decent amount of `health-apis` specific properties in here (all commented) that could be removed.

PR #8, opened 2019-05-03
CCE-13 add facilities that meet access standards to response

jblefkove-va:
https://vasdvp.atlassian.net/browse/CCE-13

Add facilities that meet access standards to community-care-eligibility response

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
@@ -137,8 +137,11 @@ public class CommunityCareEligibilityV1ApiController {
   }
 
   private CommunityCareEligibilityResponse.CommunityCareEligibility communityCareEligibility(
-      Boolean communityCareEligible, String communityCareDescription) {
-    if (allBlank(communityCareDescription, communityCareEligible)) {
+      Boolean communityCareEligible,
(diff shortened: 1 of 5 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -136,9 +136,24 @@ private static String state(VaFacilitiesResponse.Facility vaFacility) {
...
+  }
+
   private CommunityCareEligibilityResponse.CommunityCareEligibility communityCareEligibility(
-      Boolean communityCareEligible, String communityCareDescription) {
-    if (allBlank(communityCareDescription, communityCareEligible)) {
+      Boolean communityCareEligible,
+      List<String> eligibilityDescriptions,
+      List<Facility> facilitiesMeetingAccessStandards) {
+    String communityCareDescription = String.join(", ", eligibilityDescriptions);
+    if (allBlank(communityCareEligible, communityCareDescription)) {
```
jjennings-va: I don't think communityCareEligible or communityCareDescription can ever be blank. Descrition will either get a description from E&E or will get set to "Access-Standards" 

I guess it doesn't hurt anything to do a null check, just an observance. (the check below for description length is probably over kill now as well.)

jblefkove-va: You are correct, I have simplified it.

PR #9, opened 2019-05-07
Update to use Facilities search by state option

jjennings-va:
Facilities api is now called using state in place of coordinates.
API now returns every facility within the state, not just first 30
Updated swagger to properly reflect version

https://vasdvp.atlassian.net/browse/CCE-7

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
@@ -117,25 +117,6 @@ public class CommunityCareEligibilityV1ApiController {
     return map;
   }
 
-  static String state(VaFacilitiesResponse.Facility vaFacility) {
-    if (vaFacility == null) {
-      return null;
-    }
-    Attributes attributes = vaFacility.attributes();
-    if (attributes == null) {
-      return null;
-    }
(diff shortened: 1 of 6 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV1ApiController.java
```diff
@@ -266,27 +267,22 @@ public CommunityCareEligibilityResponse search(
...
-    List<VaFacilitiesResponse.Facility> filteredByServiceTypeAndState =
+    VaFacilitiesResponse vaFacilitiesResponse = facilitiesClient.facilities(patientAddress.state());
+    List<VaFacilitiesResponse.Facility> filteredByServiceType =
         vaFacilitiesResponse == null
             ? Collections.emptyList()
             : vaFacilitiesResponse
                 .data()
                 .stream()
                 .filter(vaFacility -> hasServiceType(vaFacility, filteringServiceType))
-                .filter(vaFacility -> equalsIgnoreCase(state(vaFacility), patientAddress.state()))
```
jblefkove-va: I think you can remove the `state(facility)` method too, since this was the only place it was used.

jjennings-va: Good call, oversight on my part.

PR #17, opened 2019-05-21
Remove 'api' from path

jblefkove-va:
https://vasdvp.atlassian.net/browse/CCE-34

Path clean-up goodness.


bschofield-va:
This doesn't affect any tests?

jblefkove-va:
We don't have any integration tests (yet)

PR #19, opened 2019-05-23
Updated to use Facilities API drivetime

jjennings-va:
**THIS USES MOCK DATA FOR THE FACILITIES 'NEARBY' SEARCH**

https://vasdvp.atlassian.net/browse/CCE-5

Now makes 2 calls to Facilities.

1st Call to Get Nearby Facilities (Then process response for AccessStandards)
2nd Call to get every Facility in the state (not sorted in any way)

Also includes Bing maps removal
https://vasdvp.atlassian.net/browse/CCE-4


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/RestFacilitiesClient.java
```diff
@@ -57,4 +58,34 @@ public VaFacilitiesResponse facilities(String state) {
...
+            .queryParam("per_page", 500)
+            .toUriString();
+    HttpHeaders headers = new HttpHeaders();
+    headers.add("apiKey", vaFacilitiesApiKey);
+    headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
+    VaFacilitiesResponse responseObject;
+    try {
+      responseObject =
+          restTemplate
+              .exchange(url, HttpMethod.GET, new HttpEntity<>(headers), VaFacilitiesResponse.class)
```
jblefkove-va: Because the structure of the `facility` object in the facilities API has changed for the `v1/nearby` endpoint, this should not be mapped to `VaFacilitiesResponse`. (Wait times moved and are no longer getting populated. The other fields are ok.)

Rather than set up a separate hierarchy of classes, let's use a simplified response object that only maps the facility ID:

```
class VaFacilitiesResponseV1 {
  List<Facility> data;
  class Facility { private String id; }
}
```

Have this method extract the ids from that response object and return a list of Strings representing those IDs. Then, in the main controller method, retrieve all the facilities for the state and filter that by service type. Then filter that list again to contain only the entries whose ID is also in the list of IDs from `nearby()`, then finally filter that by the wait times. That will be the list of facilities that meet the access standards and go in `CommunityCareEligibility.facilities`.

jblefkove-va:
@dbrown-va Please review as well before we merge

PR #22, opened 2019-05-29
CCE-51 Add architecture diagrams

jblefkove-va:
https://vasdvp.atlassian.net/browse/CCE-51


aparcel-va:
These diagrams show VSS as the source system for wait time information.  Should this be VIA?

jblefkove-va:
Yes it should be VIA. Is the source of these diagrams available somewhere I can edit them?

PR #27, opened 2019-06-07
Added version portion of path to controller instead of ingress

dbrown-va:
JIRA Issue: https://vasdvp.atlassian.net/browse/CCE-80

Moving the version information down to the app controller level from the ingress level allowing for greater flexibility in releases (similar to earlier Mock-EE changes)

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -142,7 +142,7 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
   /** Compute community care eligibility. */
   @Override
   @SneakyThrows
-  @GetMapping(value = "/search")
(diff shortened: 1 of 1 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -142,7 +142,7 @@ private boolean eligbleByEligbilityAndEnrollmentResponse(
   /** Compute community care eligibility. */
   @Override
   @SneakyThrows
-  @GetMapping(value = "/search")
+  @GetMapping(value = "/v0/eligibility/search")
```
jblefkove-va: It doesn't really matter since we only have one operation, but `/v0/eligibility` should move up to the controller class level.

dbrown-va: Good point.  Especially with the controller being named `CommunityCareEligibilityV0ApiController`.

I have updated to have `/v0/eligibility` as part of the controller `@RequestMapping`

PR #32, opened 2019-06-11
removed post call filtering and added service to Facilities call

jjennings-va:
Removed post Facilities call filtering and added serviceType to facilities API call. 

Code Coverage dropped 1% due to drop in total lines of code.


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -187,37 +170,28 @@ public CommunityCareEligibilityResponse search(
...
+        facilitiesClient.facilities(patientAddress.state(), mappedServiceType);
+
+    List<Facility> facilitiesInStateForService =
         stateResponse == null
             ? Collections.emptyList()
             : stateResponse
                 .data()
                 .stream()
-                .filter(vaFacility -> hasServiceType(vaFacility, mappedServiceType))
+                .collect(Collectors.toList())
```
jblefkove-va: This `collect()` in the middle of the stream looks like a copy paste error. You can remove this and the subsequent `stream()` call

PR #33, opened 2019-06-11
Task/cce 71 steel thread health check

dbrown-va:
JIRA Issue: https://vasdvp.atlassian.net/browse/CCE-71

Added a steel thread health check that verifies both E&E and Facilities API's.  Also adds corresponding tests to maintain coverage.

**Note**: Will require update to DU to include new properties.


community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -37,7 +37,7 @@
   @Data
   @Builder
   @NoArgsConstructor(access = AccessLevel.PRIVATE)
-  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PUBLIC)
```
jblefkove-va: Let's keep this private and use the builder

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -3,15 +3,15 @@
 import lombok.experimental.UtilityClass;
 
 @UtilityClass
-final class Exceptions {
-  static final class EeUnavailableException extends RuntimeException {
-    EeUnavailableException(Throwable cause) {
+public final class Exceptions {
```
jblefkove-va: CCE is a small enough app that it doesn't really need multiple packages. Put the health check class in the main package with everything else, then these can stay package-private.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/healthcheck/SteelThreadSystemCheck.java
```diff
@@ -0,0 +1,74 @@
...
+  /** 'By hand' all args constructor is required to inject non-string values from our properties. */
+  public SteelThreadSystemCheck(
+      @Autowired EligibilityAndEnrollmentClient eeClient,
+      @Autowired FacilitiesClient facilitiesClient,
+      @Value("${health-check.icn}") String icn,
+      @Value("${health-check.address.street}") String street,
+      @Value("${health-check.address.city}") String city,
+      @Value("${health-check.address.state}") String state,
+      @Value("${health-check.address.zip}") String zip,
+      @Value("${health-check.drive-minutes}") int driveMinutes) {
```
jblefkove-va: I'm inclined to hard-code the address and drive-minutes values and let `icn` be the only thing that needs configuration.

community-care-eligibility/src/main/resources/application.properties
```diff
@@ -21,3 +21,10 @@ ee.header.username=unset
 ee.request.name=CommunityCareInfo
 ee.truststore.password=unset
 ee.truststore.path=unset
+
+health-check.icn=skip
+health-check.address.street=46 W New Haven Ave
```
jblefkove-va: FYI there's not anything significant about this address... I happened to use it for testing one day and it stuck. But I would suggest good ole' `505 N John Rodes Blvd, Melbourne, FL 32934` for this

PR #38, opened 2019-06-12
changed root path for swagger

jjennings-va:
Updated root path for swagger to be 'v0/eligibility'

This PR will require a follow up change on the CCE kong route.

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HomeController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HomeController.java
@@ -26,7 +26,7 @@ public class HomeController {
 
   /** REST endpoint for OpenAPI JSON + redirect. */
   @GetMapping(
-    value = {"/", "/openapi.json", "/api/openapi.json"},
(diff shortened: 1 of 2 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HomeController.java
```diff
@@ -26,7 +26,7 @@ public HomeController(@Value("classpath:/openapi.yaml") Resource openapi) {
 
   /** REST endpoint for OpenAPI JSON + redirect. */
   @GetMapping(
-    value = {"/", "/openapi.json", "/api/openapi.json"},
+    value = {"/v0/eligibility", "/openapi.json", "/api/openapi.json"},
```
jblefkove-va: The `v0/eligibility` should go up on the controller so that the `openapi.yaml` paths are also affected. (Also this class should be renamed, `HomeControllerV0` ?)

dtournour-va: Wouldn't you also need to add /v0/eligibility/openapi.json ?

jblefkove-va:
@dtournour-va I didn't understand your last comment. Are you good with this version?

PR #41, opened 2019-06-13
Updated to short circtuit calls and responses.

jjennings-va:
https://vasdvp.atlassian.net/browse/CCE-72

Short circuits added based off EE and /nearby responses

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -24,6 +24,7 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.util.CollectionUtils;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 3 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -170,9 +157,48 @@ public CommunityCareEligibilityResponse search(
             .zip(zip.trim())
             .build();
 
+    boolean communityCareEligible = eligbleByEligbilityAndEnrollmentResponse(codeString);
+
+    if (communityCareEligible || codeString.contains("X")) {
```
jblefkove-va: Now that the `eligbleByEligbilityAndEnrollmentResponse` method has simplified to just check if the codeStrings contain G/H/N, it feels like that should be inlined here with the check for X. 

And you can use `CollectionUtils.containsAny` to simplify it a little more: `CollectionUtils.containsAny(codeStrings, Arrays.asList("G","H","N","X"))`

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -170,9 +157,48 @@ public CommunityCareEligibilityResponse search(
...
+      return CommunityCareEligibilityResponse.builder()
+          .patientRequest(
+              CommunityCareEligibilityResponse.PatientRequest.builder()
+                  .serviceType(mappedServiceType)
```
jblefkove-va: This block of code to create the request object is duplicated. Can we do this once at the beginning of the method?

PR #42, opened 2019-06-14
CCE-104 Updated hand-of-queen-elizabeth with better authentication error messages

jblefkove-va:
https://vasdvp.atlassian.net/browse/CCE-104

Merge after https://github.com/department-of-veterans-affairs/health-apis-eligibility-and-enrollment-client/pull/13


jblefkove-va:
This is ready to merge.

PR #43, opened 2019-06-18
updated format and content of output to better represent data to users.

jjennings-va:
https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=115&projectKey=CCE&view=planning&selectedIssue=CCE-94

Added new fields and modified existing fields to provide a better output for the end user.


community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -20,12 +20,37 @@
...
   PatientRequest patientRequest;
 
-  CommunityCareEligibility communityCareEligibility;
+  List<EligibilityCode> eligibilityCodes;
+
+  Boolean grandfathered;
+
+  Boolean noFullServiceVaMedicalFacility;
 
   List<Facility> facilities;
```
jjennings-va: I forgot to change this to "nearbyFacilities" 

I will do that now with all corresponding changes

jblefkove-va: 👍

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -214,16 +213,13 @@ public CommunityCareEligibilityResponse search(
...
-                .facilities(
-                    facilitiesMeetingAccessStandards
-                        .stream()
-                        .map(facility -> facility.id())
-                        .collect(Collectors.toList()))
-                .build())
-        .facilities(facilitiesInStateForService);
+        .eligible(facilitiesMeetingAccessStandards.isEmpty())
+        .eligibilityCodes(eligibilityCodes)
+        .nearbyFacilities(facilitiesInStateForService)
```
jblefkove-va: The `nearbyFacilities` should only be the results from the `nearby` call, not all of the facilities in the state.

Since we're no longer concerned with gathering all the facilities in the state, we can remove the call to `facilities(state, serviceType)` and simplify how we interact with Facilities API. That is, remove `facilities(state, serviceType)` from the client and update `facilitiesClient.nearby(patientAddress, driveMins, mappedServiceType)` to return a full `VaFacilitiesResponse`. Repurpsoe `VaFacilitiesResponse` to match the output of the `nearby` call, since the wait times have moved around.

Then the algorithm should be:
* Get nearby facilities and filter only to facilities in the patient's state. If there are none left, return.
* Transform the facilities to our model.
* Filter to facilities that meet the wait-time standard and return.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -157,34 +158,41 @@ public CommunityCareEligibilityResponse search(
...
+
+      return communityCareEligibilityResponse
+          .eligible(!codeString.contains("X"))
+          .grandfathered(codeString.contains("G"))
+          .noFullServiceVaMedicalFacility(codeString.contains("N"));
     }
 
     boolean isPrimary = equalsIgnoreCase(mappedServiceType, "primarycare");
     final int driveMins = isPrimary ? maxDriveMinsPrimary : maxDriveMinsSpecialty;
+    VaFacilitiesResponse facilityIdsWithinDriveTimeResponse =
```
jblefkove-va: Some of these names were wordy before to avoid confusion between the state call and the nearby call. Now that it's just  nearby, I think we can simplify these... how about `nearbyResponse`?

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -157,34 +158,41 @@ public CommunityCareEligibilityResponse search(
...
+        facilitiesClient.nearbyFacilities(patientAddress, driveMins, mappedServiceType);
+
     List<String> facilityIdsWithinDriveTime =
```
jblefkove-va: Since we don't have the state call any more, you don't need this list of nearby IDs.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -195,35 +203,16 @@ public CommunityCareEligibilityResponse search(
...
-                .eligibilityCode(eligibilityCodes)
-                .facilities(
-                    facilitiesMeetingAccessStandards
-                        .stream()
-                        .map(facility -> facility.id())
-                        .collect(Collectors.toList()))
-                .build())
-        .facilities(facilitiesInStateForService);
+        .eligible(facilitiesMeetingAccessStandards.isEmpty())
+        .nearbyFacilities(nearbyFacilities)
```
jblefkove-va: Let's move this up and populate the nearby facilities on the response as soon as they're calculated.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -157,34 +158,41 @@ public CommunityCareEligibilityResponse search(
...
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
     if (facilityIdsWithinDriveTime.isEmpty()) {
-      return communityCareEligibilityResponse.communityCareEligibility(
-          CommunityCareEligibilityResponse.CommunityCareEligibility.builder()
-              .eligible(true)
-              .eligibilityCode(eligibilityCodes)
-              .build());
+      return communityCareEligibilityResponse.eligible(true);
```
jblefkove-va: If there are no nearby facilities, we should indicate that on the response with an empty list instead of null. To avoid setting nearby facilities on the response more than once, I'm thinking this:

```
    List<Facility> nearbyFacilities = ...
    communityCareEligibilityResponse.nearbyFacilities(nearbyFacilities);
    if (nearbyFacilities.isEmpty()) {
      return communityCareEligibilityResponse.eligible(true);
    }
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
jjennings-va: Agreed, I will make the changes.

PR #48, opened 2019-06-27
CCE-108 Removed address param from CCE API call, retrieved from E&E

CameronPH:
JIRA: https://vasdvp.atlassian.net/browse/CCE-108

Removed the need for address parameters when querying a patient's information. 

For the `smoke-tests` and `regression-tests`, I was unable get the docker test to pass the `Token validation` tests. Derek Brown suggested that a visual verification would be sufficient in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ facilities that satisfy the access standards.
 Sample request:
 
 ```
-https://foo.com/community-care/v0/eligibility/search?patient=011235813V213455&street=742%20Evergreen%20Terrace&city=Springfield&state=KY&zip=89144&serviceType=primarycare
+https://foo.com/community-care/v0/eligibility/search?patient=011235813V213455&serviceType=primarycare
 ```
 
 Sample response:
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ For details about building and running this application, see the [developer guid
 
 ----
 
-The API supports a search query that accepts a patient ICN, the patient's home address,
+The API supports a search query that accepts a patient ICN
 and the patient's desired medical service type.
 
 The medical service type is one of:
--- a/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
+++ b/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
@@ -23,13 +23,13 @@ public final class CommunityCareEligibilityResponse {
 
   PatientRequest patientRequest;
 
-  Address patientAddress;
-
   List<EligibilityCode> eligibilityCodes;
 
   Boolean grandfathered;
 
   Boolean noFullServiceVaMedicalFacility;
+
+  Address patientAddress;
 
   List<Facility> nearbyFacilities;
 
--- a/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
+++ b/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
@@ -23,6 +23,8 @@ public final class CommunityCareEligibilityResponse {
 
   PatientRequest patientRequest;
 
+  Address patientAddress;
+
   List<EligibilityCode> eligibilityCodes;
 
   Boolean grandfathered;
@@ -129,8 +131,6 @@ public final class CommunityCareEligibilityResponse {
   public static final class PatientRequest {
 
     String patientIcn;
-
-    Address patientAddress;
 
     String serviceType;
 
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -72,19 +72,20 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
   private static Address parsePatientAddress(AddressInfo addressInfo) {
     String zip = trimToEmpty(addressInfo.getZipCode());
     if (zip.isEmpty()) {
-      if (addressInfo.getPostalCode().isEmpty()) {
+      if (trimToEmpty(addressInfo.getPostalCode()).isEmpty()) {
         zip = trimToEmpty(addressInfo.getZipcode());
+      } else {
+        zip = trimToEmpty(addressInfo.getPostalCode());
       }
-      zip = trimToEmpty(addressInfo.getPostalCode());
     }
-
-    if (addressInfo.getZipPlus4() != null && !zip.isEmpty()) {
-      zip = zip + "-" + trimToEmpty(addressInfo.getZipPlus4());
+    String zipPlus4 = trimToEmpty(addressInfo.getZipPlus4());
+    if (!zip.isEmpty() && !zipPlus4.isEmpty()) {
+      zip = zip + "-" + zipPlus4;
     }
     Address patientAddress =
         Address.builder()
             .city(trimToEmpty(addressInfo.getCity()))
-            .state(trimToEmpty(addressInfo.getState().toUpperCase()))
+            .state(trimToEmpty(addressInfo.getState()).toUpperCase())
             .street(
                 trimToEmpty(
                     trimToEmpty(addressInfo.getLine1())
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -22,6 +22,7 @@ import javax.validation.constraints.NotBlank;
 import lombok.Builder;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.util.CollectionUtils;
@@ -66,6 +67,64 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
     this.facilitiesClient = facilitiesClient;
   }
 
+  @SneakyThrows
+  private static Address parsePatientAddress(AddressInfo addressInfo) {
+    if (addressInfo == null) {
+      throw new Exceptions.MissingResidentialAddressException();
+    }
+    Address patientAddress =
+        Address.builder()
+            .city(StringUtils.trim(addressInfo.getCity()))
+            .state(StringUtils.trim(addressInfo.getState().toUpperCase()))
+            .street(
+                StringUtils.trim(
+                    StringUtils.trimToEmpty(addressInfo.getLine1())
+                        + " "
+                        + StringUtils.trimToEmpty(addressInfo.getLine2())
+                        + " "
+                        + StringUtils.trimToEmpty(addressInfo.getLine3())))
+            .build();
+    if (addressInfo.getZipCode().isEmpty()) {
+      if (addressInfo.getPostalCode().isEmpty()) {
+        patientAddress.zip(StringUtils.trimToEmpty(addressInfo.getZipCode()));
+      } else {
+        patientAddress.zip(StringUtils.trimToEmpty(addressInfo.getPostalCode()));
+      }
+    } else {
+      patientAddress.zip(StringUtils.trimToEmpty(addressInfo.getZipCode()));
+    }
+    if (addressInfo.getZipPlus4() != null) {
+      patientAddress.zip(
+          StringUtils.trimToEmpty(patientAddress.zip() + "-" + addressInfo.getZipPlus4()));
+    }
+    return patientAddress;
+  }
+
+  private static AddressInfo patientResidentialAddressInfo(GetEESummaryResponse response) {
+    if (response == null
+        || response.getSummary() == null
+        || response.getSummary().getDemographics() == null
+        || response.getSummary().getDemographics().getContactInfo() == null
+        || response.getSummary().getDemographics().getContactInfo().getAddresses() == null) {
+      return null;
+    }
+    for (AddressInfo info :
+        response.getSummary().getDemographics().getContactInfo().getAddresses().getAddress()) {
+      if ("Residential".equals(info.getAddressTypeCode())) {
+        if (info.getCity() == null
+            || info.getState() == null
+            || ((info.getZipCode() == null) && (info.getPostalCode()) == null)
+            || ((info.getLine1() == null)
+                && (info.getLine2() == null)
+                && (info.getLine3() == null))) {
+          throw new Exceptions.MissingAddressInformationException();
+        }
+        return info;
+      }
+    }
+    return null;
+  }
+
   private static Map<String, String> servicesMap() {
     Map<String, String> map = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     for (String service :
@@ -104,47 +163,6 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
         .collect(Collectors.toList());
   }
 
-  private Address parsePatientAddress(AddressInfo patientResidentialAddressInfo) {
-    Address patientAddress =
-        Address.builder()
-            .city(patientResidentialAddressInfo.getCity())
-            .state(patientResidentialAddressInfo.getState())
-            .zip(
-                (new StringBuilder()).append(patientResidentialAddressInfo.getZipCode()).toString())
-            .build();
-    StringBuilder streetBuilder = new StringBuilder();
-    for (String line :
-        Arrays.asList(
-            patientResidentialAddressInfo.getLine1(),
-            patientResidentialAddressInfo.getLine2(),
-            patientResidentialAddressInfo.getLine3())) {
-      if ((streetBuilder.length() != 0) && !line.isEmpty()) {
-        streetBuilder.append(" ");
-      }
-      streetBuilder.append(line);
-    }
-    patientAddress.street(streetBuilder.toString());
-    if (patientResidentialAddressInfo.getZipPlus4() != null) {
-      patientAddress.zip(
-          (new StringBuilder())
-              .append(patientResidentialAddressInfo.getZipCode())
-              .append("-")
-              .append(patientResidentialAddressInfo.getZipPlus4())
-              .toString());
-    }
-    return patientAddress;
-  }
-
-  private AddressInfo patientResidentialAddressInfo(GetEESummaryResponse response) {
-    for (AddressInfo info :
-        response.getSummary().getDemographics().getContactInfo().getAddresses().getAddress()) {
-      if ("Residential".equals(info.getAddressTypeCode())) {
-        return info;
-      }
-    }
-    return null;
-  }
-
   /** Compute community care eligibility. */
   @Override
   @SneakyThrows
@@ -157,8 +175,8 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
       throw new Exceptions.UnknownServiceTypeException(serviceType);
     }
     Instant timestamp = Instant.now();
-    GetEESummaryResponse response = eeClient.requestEligibility(patientIcn.trim());
-    List<VceEligibilityInfo> vceEligibilityCollection = eligibilityInfos(response);
+    GetEESummaryResponse eeResponse = eeClient.requestEligibility(patientIcn.trim());
+    List<VceEligibilityInfo> vceEligibilityCollection = eligibilityInfos(eeResponse);
     List<CommunityCareEligibilityResponse.EligibilityCode> eligibilityCodes =
         vceEligibilityCollection
             .stream()
@@ -176,14 +194,12 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
     for (int i = 0; i < eligibilityCodes.size(); i++) {
       codeString.add(eligibilityCodes.get(i).code());
     }
-    Address patientAddress = parsePatientAddress(patientResidentialAddressInfo(response));
     CommunityCareEligibilityResponse communityCareEligibilityResponse =
         CommunityCareEligibilityResponse.builder()
             .patientRequest(
                 CommunityCareEligibilityResponse.PatientRequest.builder()
                     .serviceType(mappedServiceType)
                     .patientIcn(patientIcn)
-                    .patientAddress(patientAddress)
                     .timestamp(timestamp.toString())
                     .build())
             .eligibilityCodes(eligibilityCodes)
@@ -196,6 +212,8 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
           .grandfathered(codeString.contains("G"))
           .noFullServiceVaMedicalFacility(codeString.contains("N"));
     }
+    Address patientAddress = parsePatientAddress(patientResidentialAddressInfo(eeResponse));
+    communityCareEligibilityResponse.patientAddress(patientAddress);
     boolean isPrimary = equalsIgnoreCase(mappedServiceType, "primarycare");
     final int driveMins = isPrimary ? maxDriveMinsPrimary : maxDriveMinsSpecialty;
     VaFacilitiesResponse nearbyResponse =
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -1,6 +1,7 @@
 package gov.va.api.health.communitycareeligibility.service;
 
 import static org.apache.commons.lang3.StringUtils.equalsIgnoreCase;
+import static org.apache.commons.lang3.StringUtils.trimToEmpty;
 
 import gov.va.api.health.communitycareeligibility.api.CommunityCareEligibilityResponse;
 import gov.va.api.health.communitycareeligibility.api.CommunityCareEligibilityResponse.Address;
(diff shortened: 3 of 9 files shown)


jblefkove-va:
The `README` should also be updated

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityTest.java
```diff
@@ -76,18 +82,45 @@ public void audiology() {
...
+                .summary(
+                    EeSummary.builder()
+                        .demographics(
+                            DemographicInfo.builder()
+                                .contactInfo(
+                                    ContactInfo.builder()
+                                        .addresses(
+                                            AddressCollection.builder()
+                                                .address(
+                                                    new ArrayList<AddressInfo>(
```
jblefkove-va: This `new ArrayList` wrapper is unnecessary, `Arrays.asList` already returns a list.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -161,20 +190,16 @@ public CommunityCareEligibilityResponse search(
...
             .build();
-
     if (CollectionUtils.containsAny(codeString, Arrays.asList("G", "N", "H", "X"))) {
-
       return communityCareEligibilityResponse
           .eligible(!codeString.contains("X"))
           .grandfathered(codeString.contains("G"))
           .noFullServiceVaMedicalFacility(codeString.contains("N"));
     }
-
```
jblefkove-va: The order of fields in the response should roughhly reflect the order in which the algorithm uses them. Now that address is no longer part of the request, that suggests a couple things:

* `patientAddress` should move out of `patientRequest` to the top level of the response

* `patientAddress` should appear in the response after `noFullServiceVaMedicalFacility`, because we don't need to look at the address in cases where the veteran has an E&E code that already makes them explicitly eligible or ineligible. 

* The computation of the address should move down to this line.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -140,14 +176,7 @@ public CommunityCareEligibilityResponse search(
...
-    Address patientAddress =
-        Address.builder()
-            .street(street.trim())
-            .city(city.trim())
-            .state(state.toUpperCase(Locale.US).trim())
-            .zip(zip.trim())
-            .build();
+    Address patientAddress = parsePatientAddress(patientResidentialAddressInfo(response));
```
jblefkove-va: `parsePatientAddress` throws a null pointer exception if no address is found in E&E. We should handle that more gracefully. Let's add a new exception type to throw if no residential address is found. We should be similarly restrictive for the case where the E&E address is available, but incomplete. (Missing any of the four fields.)

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+              .append("-")
+              .append(patientResidentialAddressInfo.getZipPlus4())
+              .toString());
+    }
+    return patientAddress;
+  }
+
+  private AddressInfo patientResidentialAddressInfo(GetEESummaryResponse response) {
+    for (AddressInfo info :
+        response.getSummary().getDemographics().getContactInfo().getAddresses().getAddress()) {
```
jblefkove-va: In general, we need to be cautious when dealing with data from other APIs. There are no guarantees about availability, so any of the calls in this long chain could potentially be null, and we should handle that. Take a look at the other method in this class that extracts the eligibility codes from the eeResponse.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+            .state(patientResidentialAddressInfo.getState())
+            .zip(
+                (new StringBuilder()).append(patientResidentialAddressInfo.getZipCode()).toString())
+            .build();
+    StringBuilder streetBuilder = new StringBuilder();
+    for (String line :
+        Arrays.asList(
+            patientResidentialAddressInfo.getLine1(),
+            patientResidentialAddressInfo.getLine2(),
+            patientResidentialAddressInfo.getLine3())) {
```
jblefkove-va: This could be done more elegantly with a stream using `Collectors.join(", ")`. (i.e. create a stream with the three lines, trim  them, discard the blanks, then concatenate)

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
         .collect(Collectors.toList());
   }
 
+  private Address parsePatientAddress(AddressInfo patientResidentialAddressInfo) {
+    Address patientAddress =
+        Address.builder()
+            .city(patientResidentialAddressInfo.getCity())
+            .state(patientResidentialAddressInfo.getState())
```
jblefkove-va: State should be shifted to uppercase.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
         .collect(Collectors.toList());
   }
 
+  private Address parsePatientAddress(AddressInfo patientResidentialAddressInfo) {
+    Address patientAddress =
+        Address.builder()
+            .city(patientResidentialAddressInfo.getCity())
```
jblefkove-va: As a good practice, these strings from the other API should be trimmed. Take a look at some of the methods in the FacilityTransformer class.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+            .state(patientResidentialAddressInfo.getState())
+            .zip(
+                (new StringBuilder()).append(patientResidentialAddressInfo.getZipCode()).toString())
```
jblefkove-va: You don't need a stringbuilder to only append one thing to it.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+            patientResidentialAddressInfo.getLine1(),
+            patientResidentialAddressInfo.getLine2(),
+            patientResidentialAddressInfo.getLine3())) {
+      if ((streetBuilder.length() != 0) && !line.isEmpty()) {
+        streetBuilder.append(" ");
+      }
+      streetBuilder.append(line);
+    }
+    patientAddress.street(streetBuilder.toString());
+    if (patientResidentialAddressInfo.getZipPlus4() != null) {
```
jblefkove-va: All of the logic for computing the zip code, including the extension, should be together.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+          (new StringBuilder())
+              .append(patientResidentialAddressInfo.getZipCode())
+              .append("-")
+              .append(patientResidentialAddressInfo.getZipPlus4())
+              .toString());
+    }
+    return patientAddress;
+  }
+
+  private AddressInfo patientResidentialAddressInfo(GetEESummaryResponse response) {
```
jblefkove-va: This method and the one above can be static.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
         .collect(Collectors.toList());
   }
 
+  private Address parsePatientAddress(AddressInfo patientResidentialAddressInfo) {
```
jblefkove-va: The parameter name is unnecessarily long and makes things a little harder to read. In the context of this method, we're just converting some EE address to our model of address. It isn't specially for a patient, or residential.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
...
+      }
+      streetBuilder.append(line);
+    }
+    patientAddress.street(streetBuilder.toString());
+    if (patientResidentialAddressInfo.getZipPlus4() != null) {
+      patientAddress.zip(
+          (new StringBuilder())
+              .append(patientResidentialAddressInfo.getZipCode())
+              .append("-")
+              .append(patientResidentialAddressInfo.getZipPlus4())
```
jblefkove-va: StringBuilder is mainly useful for building strings in loops, or for situations where you're building a string with complicated logic with a lot of branches. It's not really necessary here.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -104,25 +104,61 @@ public CommunityCareEligibilityV0ApiController(
         .collect(Collectors.toList());
   }
 
+  private Address parsePatientAddress(AddressInfo patientResidentialAddressInfo) {
+    Address patientAddress =
+        Address.builder()
+            .city(patientResidentialAddressInfo.getCity())
+            .state(patientResidentialAddressInfo.getState())
+            .zip(
```
jblefkove-va: If `zipCode` is blank, we should use `postalCode` instead. And if that is blank, fall back to the deprecated `zipcode`.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
CameronPH: After reviewing how the street is concatenated from three different lines in `FacilityTransformer.java`, I can just use a combination of `StringUtils.trimToEmpty()` with a regular string concatenation operator.

The reason it was set up this way before was I did not know how to prevent attempting to access one of the streets if it was `null`. But `StringUtils.trimToEmpty()` converts `null` strings to the empty string.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
CameronPH: In regards to the cases where the method doesn't return an `AddressInfo` object with `addressTypeCode = "Residential"`. What should this method return? I could return a new instantiation of an `AddressInfo` object, and all fields that are set to `null` will be converted to the empty string through `StringUtils.trimToEmpty()`. So it wouldn't throw any errors and we would just get an `Address` object with all empty strings.

Is that acceptable or do you want an error message/code of some kind to be returned when no "Residential" `addressTypeCode` is found.

CameronPH: Disregard my statement, you've already stated that you want an new exception type for that exact case.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -66,6 +67,64 @@ public CommunityCareEligibilityV0ApiController(
...
+                StringUtils.trim(
+                    StringUtils.trimToEmpty(addressInfo.getLine1())
+                        + " "
+                        + StringUtils.trimToEmpty(addressInfo.getLine2())
+                        + " "
+                        + StringUtils.trimToEmpty(addressInfo.getLine3())))
+            .build();
+    if (addressInfo.getZipCode().isEmpty()) {
+      if (addressInfo.getPostalCode().isEmpty()) {
+        patientAddress.zip(StringUtils.trimToEmpty(addressInfo.getZipCode()));
```
CameronPH: You stated that if neither the zipCode nor postalCode we filled in, to fall back on the zipCode. Would it not be better throw an MissingAddressInformationException() instead?

README.md
```diff
@@ -56,7 +56,7 @@ facilities that satisfy the access standards.
 Sample request:
 
 ```
-https://foo.com/community-care/v0/eligibility/search?patient=011235813V213455&street=742%20Evergreen%20Terrace&city=Springfield&state=KY&zip=89144&serviceType=primarycare
+https://foo.com/community-care/v0/eligibility/search?patient=011235813V213455&serviceType=primarycare
```
jblefkove-va: There is a line above that mentions requesting patient home address that should also be removed. And the sample response below should be updated.

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityResponse.java
```diff
@@ -23,6 +23,8 @@
 
   PatientRequest patientRequest;
 
+  Address patientAddress;
```
jblefkove-va: The order of fields here should match the order that the algorithm uses them... so put address after `noFullServiceVaMedicalFacility`

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -27,4 +27,16 @@
       super("Unknown service type: " + serviceType);
     }
   }
+
+  static final class MissingResidentialAddressException extends NullPointerException {
```
jblefkove-va: Keep the exceptions alphabetized. And these should extend `RuntimeException`

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -27,4 +27,16 @@
...
+    MissingResidentialAddressException() {
+      super("No residential address found");
+    }
+  }
+
+  static final class MissingAddressInformationException extends NullPointerException {
+    MissingAddressInformationException() {
+      super("Residential address has incomplete information");
```
jblefkove-va: To help with debugging, the incomplete address should be included on the exception. For example, if only the zip code is missing, showing the address object with street/city/state populated is helpful. So I would move the check for missing data to the end of the `parsePatientAddress` method, and throw the `MissingAddressInformationException` at that time

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -27,4 +27,16 @@
       super("Unknown service type: " + serviceType);
     }
   }
+
+  static final class MissingResidentialAddressException extends NullPointerException {
+    MissingResidentialAddressException() {
+      super("No residential address found");
```
jblefkove-va: Include the patient ICN in this error message.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -27,4 +27,16 @@
...
+    }
+  }
+
+  static final class MissingAddressInformationException extends NullPointerException {
```
jblefkove-va: At a glance `MissingResidentialAddress` and `MissingAddressInformation` are easy to get mixed up. How about calling this `IncompleteAddressException`.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,70 @@ public CommunityCareEligibilityV0ApiController(
...
   }
 
+  @SneakyThrows
+  private static Address parsePatientAddress(AddressInfo addressInfo) {
+    if (addressInfo == null) {
+      throw new Exceptions.MissingResidentialAddressException();
+    }
+    Address patientAddress =
+        Address.builder()
+            .city(StringUtils.trim(addressInfo.getCity()))
```
jblefkove-va: Use static imports to make all these calls to `StringUtils` methods more readable.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,67 @@ public CommunityCareEligibilityV0ApiController(
...
+        || response.getSummary() == null
+        || response.getSummary().getDemographics() == null
+        || response.getSummary().getDemographics().getContactInfo() == null
+        || response.getSummary().getDemographics().getContactInfo().getAddresses() == null) {
+      return null;
+    }
+    for (AddressInfo info :
+        response.getSummary().getDemographics().getContactInfo().getAddresses().getAddress()) {
+      if ("Residential".equals(info.getAddressTypeCode())) {
+        return info;
```
jblefkove-va: This pattern of finding the first matching item in a list can be done more elegantly with a stream. Take a look at https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/crawler/ResourceDiscovery.java#L112

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/Exceptions.java
```diff
@@ -16,6 +17,31 @@
     }
   }
 
+  static final class IncompleteAddressException extends RuntimeException {
+    IncompleteAddressException(Address patientAddress) {
+      super(
+          "Residential address has incomplete information"
```
jblefkove-va: The extra formatting isn't really necessary. You can do `"Residential address has incomplete information: " + patientAddress` and rely on the address `toString` method.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,67 @@ public CommunityCareEligibilityV0ApiController(
...
+    if (addressInfo.getZipCode().isEmpty()) {
+      if (addressInfo.getPostalCode().isEmpty()) {
+        patientAddress.zip(trimToEmpty(addressInfo.getZipCode()));
+      } else {
+        patientAddress.zip(trimToEmpty(addressInfo.getPostalCode()));
+      }
+    } else {
+      patientAddress.zip(trimToEmpty(addressInfo.getZipCode()));
+    }
+    if (addressInfo.getZipPlus4() != null) {
```
jblefkove-va: There should be a check that the extension is only added if the base zip code is not blank. Otherwise `-1234` will look like a valid zip code, but is not.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,67 @@ public CommunityCareEligibilityV0ApiController(
...
+                trimToEmpty(
+                    trimToEmpty(addressInfo.getLine1())
+                        + " "
+                        + trimToEmpty(addressInfo.getLine2())
+                        + " "
+                        + trimToEmpty(addressInfo.getLine3())))
+            .build();
+    if (addressInfo.getZipCode().isEmpty()) {
+      if (addressInfo.getPostalCode().isEmpty()) {
+        patientAddress.zip(trimToEmpty(addressInfo.getZipCode()));
```
jblefkove-va: If `zipCode` and `postalCode` are both blank, you can fall back on `zipcode` (all lowercase) as a last attempt to populate the zip code.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,67 @@ public CommunityCareEligibilityV0ApiController(
     this.facilitiesClient = facilitiesClient;
   }
 
+  @SneakyThrows
+  private static Address parsePatientAddress(AddressInfo addressInfo) {
+    Address patientAddress =
+        Address.builder()
```
jblefkove-va: It's confusing that the address is built up front, but not all of the fields are populated. Calculate the zip code first and store that in a local variable, then build the address and set all four fields at the same time.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +67,67 @@ public CommunityCareEligibilityV0ApiController(
...
+            .state(trimToEmpty(addressInfo.getState().toUpperCase()))
+            .street(
+                trimToEmpty(
+                    trimToEmpty(addressInfo.getLine1())
+                        + " "
+                        + trimToEmpty(addressInfo.getLine2())
+                        + " "
+                        + trimToEmpty(addressInfo.getLine3())))
+            .build();
+    if (addressInfo.getZipCode().isEmpty()) {
```
jblefkove-va: These checks are easier to follow if you initialize `zip` to `zipCode` and then override it if `zip` is empty:

```
String zip = trimToEmpty(addressInfo.getZipCode());
if (zip.isEmpty()) {
  zip = trimToEmpty(addressInfo.getPostalCode());
} 
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +68,73 @@ public CommunityCareEligibilityV0ApiController(
...
+      zip = trimToEmpty(addressInfo.getPostalCode());
+    }
+
+    if (addressInfo.getZipPlus4() != null && !zip.isEmpty()) {
+      zip = zip + "-" + trimToEmpty(addressInfo.getZipPlus4());
+    }
+    Address patientAddress =
+        Address.builder()
+            .city(trimToEmpty(addressInfo.getCity()))
+            .state(trimToEmpty(addressInfo.getState().toUpperCase()))
```
jblefkove-va: This is a `NullPointerException` if `getState()` is null.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +68,73 @@ public CommunityCareEligibilityV0ApiController(
     this.facilitiesClient = facilitiesClient;
   }
 
+  @SneakyThrows
+  private static Address parsePatientAddress(AddressInfo addressInfo) {
+    String zip = trimToEmpty(addressInfo.getZipCode());
+    if (zip.isEmpty()) {
+      if (addressInfo.getPostalCode().isEmpty()) {
```
jblefkove-va: This is a `NullPointerException` if `getPostalCode()` is null.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +68,73 @@ public CommunityCareEligibilityV0ApiController(
...
+        zip = trimToEmpty(addressInfo.getZipcode());
```
jblefkove-va: If this line executes to set `zip` to `Zipcode` the next line will still execute and set it to the blank `postalCode`

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -65,13 +68,73 @@ public CommunityCareEligibilityV0ApiController(
...
+      }
+      zip = trimToEmpty(addressInfo.getPostalCode());
+    }
+
+    if (addressInfo.getZipPlus4() != null && !zip.isEmpty()) {
+      zip = zip + "-" + trimToEmpty(addressInfo.getZipPlus4());
```
jblefkove-va: In the (admittedly unlikely) case that `zipPlus4` is not null, but is blank, the `zip` will be something like `12345-`, which isn't ideal.

jblefkove-va:
To test for some of the `NullPointerException`s I pointed out, I recommend adding a unit test with an address that has the type set to `Residential` and everything else `null`. And verify that the test throws the `IncompleteAddressException`.

PR #57, opened 2019-10-09
Added in new fields and updated tests

jjennings-va:
https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=115&modal=detail&selectedIssue=CCE-126

Handled new Mobile and Active fields coming from Facilities API.

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
@@ -13,7 +13,15 @@ import org.apache.commons.lang3.StringUtils;
 
 @Builder
 public class FacilityTransformer {
+
   @NonNull private final String serviceType;
+
+  private static Boolean active(VaFacilitiesResponse.Facility vaFacility) {
+    if (vaFacility.attributes() == null || vaFacility.attributes().active() == null) {
+      return false;
+    }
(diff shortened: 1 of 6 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
```diff
@@ -13,8 +13,16 @@
 
 @Builder
 public class FacilityTransformer {
+
   @NonNull private final String serviceType;
 
+  private static Boolean active(VaFacilitiesResponse.Facility vaFacility) {
+    if (vaFacility.attributes() == null || vaFacility.attributes().active() == null) {
```
jblefkove-va: The second check is unnecessary here, since `trimToNull(vaFacility.attributes().active())` will handle null

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
```diff
@@ -13,8 +13,16 @@
...
+      return false;
+    }
+    return trimToNull(vaFacility.attributes().active()).equals("A");
```
jblefkove-va: This is a NullPointer if `active()` trims to null... how about `equalsIgnoreCase(trimToNull(vaFacility.attributes().active()), "A")` instead?

jblefkove-va:
Also, update the README, and add the new fields to the Swagger example data.

PR #58, opened 2019-10-09
Task/cce 151 optional extended drive time param

jjennings-va:
https://vasdvp.atlassian.net/browse/CCE-151

Added in optional param "driveMin" if populated facility results outside of the standard eligibility radius are desired.


community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
```diff
@@ -104,5 +105,14 @@ CommunityCareEligibilityResponse search(
                 )
           )
           @NotBlank
-          String serviceType);
+          String serviceType,
+      @Parameter(
+            in = ParameterIn.QUERY,
+            name = "driveMin",
```
jblefkove-va: Having thought about this a little more, lets' call this `extendedDriveMins` and add validation in the controller that this must be greater than the base 30/60 minute value.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -188,7 +189,8 @@ private static Coordinates toCoordinates(String patientIcn, GeocodingInfo geocod
   public CommunityCareEligibilityResponse search(
       @RequestHeader(value = "X-VA-SESSIONID", defaultValue = "") String optSessionIdHeader,
       @NotBlank @RequestParam(value = "patient") String patientIcn,
-      @NotBlank @RequestParam(value = "serviceType") String serviceType) {
+      @NotBlank @RequestParam(value = "serviceType") String serviceType,
+      @Max(value = 90) @RequestParam(value = "driveMin", defaultValue = "0") int extendedDriveMin) {
```
jblefkove-va: Can we do `Integer` here instead and let it default to `null`? That makes it more explicit when it's not set.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -208,10 +209,11 @@ public CommunityCareEligibilityResponse search(
             .patientIcn(patientIcn.trim())
             .serviceType(mappedServiceType)
             .timestamp(Instant.now().toString())
-            .build());
+            .build(),
+        extendedDriveMin);
   }
 
-  private CommunityCareEligibilityResponse search(PatientRequest request) {
+  private CommunityCareEligibilityResponse search(PatientRequest request, int extendedDriveMin) {
```
jblefkove-va: Since the extended drive time is a new argument on the request, let's add it to `PatientRequest`

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityTest.java
```diff
@@ -74,6 +74,36 @@ public void audiology() {
                                     .build())
                             .build()))
                 .build());
+    when(facilitiesClient.nearbyFacilities(
+            Coordinates.builder()
+                .latitude(new BigDecimal("28.112506"))
+                .longitude(new BigDecimal("-80.7000423"))
+                .build(),
+            90,
```
jblefkove-va: In this test we get the same single facility returned for the 60 and 90 minute searches, so that doesn't really prove which one is being used to populate the nearby facilities. I suggest leaving this test unchanged and adding a new test that returns no facilities for 60 and 1 facility for 90. And the end result should be `eligible` with that one facility still included.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
jblefkove-va: Add smoke tests for this new parameter... including ones that blow up on negative numbers or > 90.

community-care-eligibility/src/main/docker/entrypoint.sh
```diff
@@ -76,6 +76,14 @@ smokeTest() {
   path="/search?serviceType=$SERVICE_TYPE&patient=$PATIENT"
   doCurl 200 $TOKEN
 
+  # extendedDriveMin parameter greater than 90
+  path="/search?serviceType=$SERVICE_TYPE&patient=$PATIENT&extendedDriveMin=91"
```
jblefkove-va: Even though we're not going to enforce it, let's make this `100` to stick with multiples of 10

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
```diff
@@ -104,5 +106,15 @@ CommunityCareEligibilityResponse search(
...
           )
           @NotBlank
-          String serviceType);
+          String serviceType,
+      @Parameter(
+            in = ParameterIn.QUERY,
+            name = "extendedDriveMin",
+            description =
+                "Optional field user provided drive radius for VA Facilities;"
+                    + " does not impact eligibility"
```
jblefkove-va: Description should indicate that this value must be greater than the configured drive time for the service type

PR #61, opened 2019-10-15
API-1538 - use queen elizabeth service

beaugrantham-va:
https://vasdvp.atlassian.net/browse/API-1538


community-care-eligibility/src/main/resources/application.properties
```diff
@@ -15,10 +15,8 @@ community-care.max-drive-time-min-primary=30
 community-care.max-drive-time-min-specialty=60
 
 va-facilities.url=unset
-ee.endpoint.url=unset
-ee.header.password=unset
-ee.header.username=unset
```
beaugrantham-va: @alecbert-va @LesleyBrownLiberty Do these need to be added back in based on the latest updates to the UC branch?

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -182,6 +182,22 @@ private static Coordinates toCoordinates(String patientIcn, GeocodingInfo geocod
...
+   *
+   * @param icn ICN to request.
+   * @return GetEESummaryResponse.
+   */
+  private GetEESummaryResponse requestEligibility(final String icn) {
+    try {
+      return eeClient.getEeSummary(icn);
+    } catch (PersonNotFound e) {
+      throw new Exceptions.UnknownPatientIcnException(icn, e);
+    } catch (EligibilitiesException e) {
```
jblefkove-va: I'd like to err on the side of wrapping any other exception from QE (besides person-not-found) as an `EeUnavailableException`

community-care-eligibility/src/main/resources/application.properties
```diff
@@ -15,10 +15,9 @@ community-care.max-drive-time-min-primary=30
...
 va-facilities.url=unset
-ee.endpoint.url=unset
+
+# Mock EE or SOAPUI endpoint url when running locally as dev.
+ee.endpoint.url=http://localhost:9090/v0/ws/
+ee.request.name=CommunityCareInfo
 ee.header.password=unset
 ee.header.username=unset
-ee.request.name=CommunityCareInfo
-ee.truststore.password=unset
```
jblefkove-va: Removing the truststore variables is going to mess up the `make-configs` script, which only overwrites properties that are `unset` in the base properties file. However, I think it is safe to remove these variables from `make-configs` as well. The `make-configs` script predates Mock-EE, so it expects real E&E url, username/password, etc. to be sourced in `secrets.conf`. Let's update `make-configs` to remove the truststore properties and hard-code Mock-EE url and credentials.

community-care-eligibility/src/main/resources/application.properties
```diff
@@ -15,10 +15,9 @@ community-care.max-drive-time-min-primary=30
 community-care.max-drive-time-min-specialty=60
 
 va-facilities.url=unset
-ee.endpoint.url=unset
+
+# Mock EE or SOAPUI endpoint url when running locally as dev.
+ee.endpoint.url=http://localhost:9090/v0/ws/
```
jblefkove-va: Similar to the comment below, don't put development values in `application.properties`. Leave this `unset` and let `make-configs` override this with `http://localhost:9090/v0/ws/` instead.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
lbrown-va: Changed to catch any Exception.

community-care-eligibility/src/main/resources/application.properties
lbrown-va: I'll just revert application.properties.  I suggest opening a different story for modifying make-configs.

community-care-eligibility/src/main/resources/application.properties
lbrown-va: I'll just revert application.properties.  I suggest opening a different story for modifying make-configs.

jblefkove-va:
I created [CCE-157](https://github.com/department-of-veterans-affairs/health-apis-community-care-eligibility/pull/62) for the `make-configs` changes, but this turned out to be complicated; the current implementation of QE requires the two trust-store parameters to be set to something, even when they are unused. So I am just going to merge that branch into this one and retest in here before this merges.

PR #65, opened 2019-10-22
Update CCE to Spring 2.2.0

jnguy-va:
https://vasdvp.atlassian.net/browse/SSS-1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>gov.va.api.health</groupId>
     <artifactId>health-apis-parent</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>gov.va.api.health</groupId>
     <artifactId>health-apis-parent</artifactId>
-    <version>3.1.9</version>
+    <version>4.0.0-SNAPSHOT</version>
```
jblefkove-va: `4.0.0` is released, so you can drop the snapshot

PR #69, opened 2019-11-13
Task/sss 9 service type not required

jjennings-va:
https://vasdvp.atlassian.net/browse/SSS-9

Allows CCE to work without serviceType provided. Short circuited check to just E&E.

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -216,7 +216,7 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
           stripNewlines(serviceType));
     }
 
-    String mappedServiceType = serviceType == null ? null : SERVICES_MAP.get(serviceType.trim());
+    String mappedServiceType = SERVICES_MAP.get(trimToEmpty(serviceType));
     if (serviceType != null && mappedServiceType == null) {
       throw new Exceptions.UnknownServiceTypeException(serviceType);
     }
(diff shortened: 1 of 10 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -269,6 +269,13 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
           .noFullServiceVaMedicalFacility(codeStrings.contains("N"))
           .build();
     }
+    if (request.serviceType() == null) {
+      return response
+          .patientRequest(request)
+          .grandfathered(false)
+          .noFullServiceVaMedicalFacility(false)
+          .build();
```
jblefkove-va: I think `return response.build()` is sufficient here, because those other fields are set on the response builder back on line 257

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -216,8 +216,8 @@ public CommunityCareEligibilityResponse search(
           stripNewlines(serviceType));
     }
 
-    String mappedServiceType = SERVICES_MAP.get(serviceType.trim());
-    if (mappedServiceType == null) {
+    String mappedServiceType = serviceType == null ? null : SERVICES_MAP.get(serviceType.trim());
```
jblefkove-va: how about `SERVICES_MAP.get(trimToEmpty(serviceType))` to avoid the extra null check

jblefkove-va:
Let's also add a unit test for this

PR #71, opened 2019-11-14
Added in pared-down facilities calls and updated tests

jjennings-va:
https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=129&view=planning&selectedIssue=SSS-22

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -181,6 +181,18 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
     return Coordinates.builder().latitude(lat).longitude(lng).build();
   }
 
+  private String convertToCommaDelimitedString(VaNearbyFacilitiesResponse nearbyResponse) {
+    String ids = "";
+    for (int i = 0; i < nearbyResponse.data().size(); i++) {
+      if (i == 0) {
+        ids += nearbyResponse.data().get(i).id();
+      } else {
+        ids += "," + nearbyResponse.data().get(i).id();
+      }
+    }
+    return ids;
+  }
+
   /**
    * Wrap the QueenElizabethService call to encapsulate any exceptions into CCE specific exceptions.
    *
@@ -216,11 +228,10 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
           stripNewlines(serviceType));
     }
 
-    String mappedServiceType = SERVICES_MAP.get(serviceType.trim());
-    if (mappedServiceType == null) {
+    String mappedServiceType = serviceType == null ? null : SERVICES_MAP.get(serviceType.trim());
+    if (serviceType != null && mappedServiceType == null) {
       throw new Exceptions.UnknownServiceTypeException(serviceType);
     }
-
     int driveMins =
         equalsIgnoreCase(mappedServiceType, "primarycare")
             ? maxDriveMinsPrimary
@@ -228,7 +239,6 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
     if (extendedDriveMin != null && extendedDriveMin <= driveMins) {
       throw new Exceptions.InvalidExtendedDriveMin(mappedServiceType, extendedDriveMin, driveMins);
     }
-
     return search(
         PatientRequest.builder()
             .patientIcn(patientIcn.trim())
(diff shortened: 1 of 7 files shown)


community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/RestFacilitiesClientTest.java
```diff
@@ -15,21 +15,38 @@
...
+    RestTemplate restTemplate = mock(RestTemplate.class);
+    when(restTemplate.exchange(
+            eq("http://foo/bar/v0/facilities?ids=vha_675GD"),
+            eq(HttpMethod.GET),
+            any(HttpEntity.class),
+            eq(VaFacilitiesResponse.class)))
+        .thenReturn(response);
+    RestFacilitiesClient client =
+        new RestFacilitiesClient("fakeApiKey", "http://foo/bar", restTemplate);
+    assertThat(client.facilitiesById("vha_675GD"))
```
jblefkove-va: Assuming that the comma-delimiting gets moved inside the rest client, let's test with a pair of IDs here to verify that they are properly comma-delimited in the URL

jblefkove-va:
`longitude` is not mapped in the response. (We're back to using the `v0` version of the full facilities object here, and I recall that they changed the name of the longitude field in `v1`.)

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -292,12 +292,19 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
...
-    VaFacilitiesResponse nearbyResponse =
+
+    VaNearbyFacilitiesResponse nearbyResponse =
         facilitiesClient.nearbyFacilities(patientCoordinates, driveMins(serviceType), serviceType);
-    List<Facility> nearbyFacilities =
+    VaFacilitiesResponse vaFacilitiesResponse =
         nearbyResponse == null
+            ? VaFacilitiesResponse.builder().build()
+            : facilitiesClient.facilitiesByIds(
+                nearbyResponse.data().stream().map(fac -> fac.id()).collect(Collectors.toList()));
```
jblefkove-va: Since the client should handle empty lists, how about:
```
facilitiesClient.facilitiesByIds(nearbyResponse == null ? emptyList() : nearbyResponse.data().stream().map(fac -> fac.id()).collect(Collectors.toList()))
```

This way the code for the empty response is contained to just the client `facilitiesByIds()` method

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/RestFacilitiesClient.java
```diff
@@ -33,7 +34,32 @@ public RestFacilitiesClient(
 
   @Override
   @SneakyThrows
-  public VaFacilitiesResponse nearbyFacilities(
+  public VaFacilitiesResponse facilitiesByIds(List<String> ids) {
```
jblefkove-va: This method should bail out with an empty response if `ids` is empty

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/RestFacilitiesClient.java
```diff
@@ -33,7 +34,32 @@ public RestFacilitiesClient(
...
+    String url =
+        UriComponentsBuilder.fromHttpUrl(baseUrl + "v0/facilities")
+            .queryParam("ids", String.join(",", ids))
+            .build()
+            .toUriString();
+    HttpHeaders headers = headers();
```
jblefkove-va: This can be inlined down below since the headers are only used in one place

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityTest.java
```diff
@@ -53,6 +53,14 @@ public void audiology() {
...
             60,
             "Audiology"))
+        .thenReturn(
+            VaNearbyFacilitiesResponse.builder()
+                .data(
+                    asList(
+                        VaNearbyFacilitiesResponse.Facility.builder().id("FAC456").build(),
+                        VaNearbyFacilitiesResponse.Facility.builder().id("FAC123").build()))
+                .build());
+    when(facilitiesClient.facilitiesByIds(asList("FAC456", "FAC123")))
```
jblefkove-va: Although this still works, it doesn't quite make sense to call `facilitiesByIds` with two IDs and only get on facility back.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
jjennings-va: I don't think I can just set it to emptyList(), since it is expecting just one VaFacilitiesResponse, not a list of Responses. 

I could refactor it to expect a list, but would it ever be more than 1?

PR #73, opened 2019-11-15
Task/sss 24 add drive time band

jjennings-va:
https://vasdvp.atlassian.net/browse/SSS-24

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.NotBlank;
@@ -187,18 +188,6 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
         : maxDriveMinsSpecialty;
   }
 
-  private VaNearbyFacilitiesResponse.Facility getCorrespondingFacility(
-      VaFacilitiesResponse.Facility vaFacility,
-      VaNearbyFacilitiesResponse nearbyFacilitiesResponse) {
-    VaNearbyFacilitiesResponse.Facility verifiedMatch = null;
-    for (int i = 0; i < nearbyFacilitiesResponse.data().size(); i++) {
-      if (vaFacility.id().equals(nearbyFacilitiesResponse.data().get(i).id())) {
-        verifiedMatch = nearbyFacilitiesResponse.data().get(i);
-      }
-    }
-    return verifiedMatch;
-  }
-
   private GetEESummaryResponse requestEligibility(final String icn) {
     try {
       return eeClient.getEeSummary(icn);
@@ -318,6 +307,14 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
       Coordinates coordinates, int driveMins, String serviceType) {
     VaNearbyFacilitiesResponse nearbyResponse =
         facilitiesClient.nearbyFacilities(coordinates, driveMins, serviceType);
+    Map<String, VaNearbyFacilitiesResponse.Facility> facilityMap =
+        nearbyResponse == null
+            ? null
+            : nearbyResponse
+                .data()
+                .stream()
+                .collect(Collectors.toMap(fac -> fac.id(), Function.identity()));
+
     VaFacilitiesResponse vaFacilitiesResponse =
         facilitiesClient.facilitiesByIds(
             nearbyResponse == null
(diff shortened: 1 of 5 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -312,15 +319,15 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
                         FacilityTransformer.builder()
                             .serviceType(serviceType)
                             .build()
-                            .toFacility(vaFacility))
+                            .toFacility(
+                                vaFacility, getCorrespondingFacility(vaFacility, nearbyResponse)))
```
jblefkove-va: Having thought about this a little more, I think the better approach here is to create a map of of each ID in `nearbyResponse` to its corresponding facility object. then do a lookup of that map here. 

You can create the map pretty cleanly with a stream:
```
        nearbyResponse
            .data()
            .stream()
            .collect(Collectors.toMap(fac -> fac.id(), Function.identity()));
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -341,7 +348,9 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
                           FacilityTransformer.builder()
                               .serviceType(serviceType)
                               .build()
-                              .toFacility(vaFacility))
+                              .toFacility(
+                                  vaFacility,
+                                  getCorrespondingFacility(vaFacility, extendedResponse)))
```
jblefkove-va: The logic for making the Facilities call and transforming the results is duplicated down here for the extended call. With the coordination of separate facilities calls, I think that has become complicated enough to warrant refactoring it out into its own method. Should be a method that takes patient coordinates, the service type, and the drive mins (regular or extended) as parameters, and returns the converted list of facilities.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
```diff
@@ -66,6 +68,17 @@ private static Coordinates coordinates(VaFacilitiesResponse.Facility vaFacility)
         .build();
   }
 
+  private static CommunityCareEligibilityResponse.DriveMinutes driveMinutes(
+      VaNearbyFacilitiesResponse.Facility vaFacility) {
```
jblefkove-va: Should this be `nearbyFacility`?

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/FacilityTransformer.java
```diff
@@ -98,7 +111,9 @@ private static String website(VaFacilitiesResponse.Facility vaFacility) {
   }
 
   /** Check for Facility. */
-  public Facility toFacility(VaFacilitiesResponse.Facility vaFacility) {
+  public Facility toFacility(
+      VaFacilitiesResponse.Facility vaFacility,
+      VaNearbyFacilitiesResponse.Facility nearbyFacility) {
     if (vaFacility == null) {
```
jblefkove-va: I think we should also bail out if `nearbyFacility` is null.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -293,58 +291,46 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
...
+            : nearbyResponse
+                .data()
+                .stream()
+                .collect(Collectors.toMap(fac -> fac.id(), Function.identity()));
+
+    VaFacilitiesResponse vaFacilitiesResponse =
+        facilitiesClient.facilitiesByIds(
+            nearbyResponse == null
+                ? Collections.emptyList()
+                : nearbyResponse.data().stream().map(fac -> fac.id()).collect(Collectors.toList()));
```
jblefkove-va: Now that this is its own method, I think it would be cleaner to pull out the null check:
```
if (nearbyResponse == null) { return emptyList(); }
 facilitiesClient.facilitiesByIds(nearbyResponse.data().stream().map(fac -> fac.id()).collect(Collectors.toList()));
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -293,58 +291,46 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
...
+        ? Collections.emptyList()
+        : vaFacilitiesResponse
+            .data()
+            .stream()
+            .map(
+                vaFacility ->
+                    FacilityTransformer.builder()
+                        .serviceType(serviceType)
+                        .build()
+                        .toFacility(vaFacility, facilityMap.get(vaFacility.id())))
```
jblefkove-va: since `toFacility()` may return `null`, let's add a last step to the stream to filter out the nulls. (I think we already do this with the stream for transforming the E&E codes earlier)

PR #83, opened 2020-02-12
SHANK-308 Upgrade to 6.0 parent with Error Prone

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-308


jblefkove-va:
This is the first (and probably last) time I've ever run a new static analysis tool against existing code with no findings!

PR #85, opened 2020-02-26
SHANK-353 Replace dev-app with run-local

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-353


jhulbert-va:
Will become part of a tools repo instead.

PR #88, opened 2020-03-02
Shank 336 remove geocoding exception

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-336

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -169,13 +169,13 @@ public class CommunityCareEligibilityV0ApiController implements CommunityCareEli
         .build();
   }
 
-  private static Coordinates toCoordinates(String patientIcn, GeocodingInfo geocodingInfo) {
+  private static Optional<Coordinates> toCoordinates(GeocodingInfo geocodingInfo) {
     BigDecimal lat = geocodingInfo.getAddressLatitude();
     BigDecimal lng = geocodingInfo.getAddressLongitude();
     if (lat == null || lng == null) {
-      throw new Exceptions.MissingGeocodingInfoException(patientIcn);
+      return Optional.empty();
     }
-    return Coordinates.builder().latitude(lat).longitude(lng).build();
+    return Optional.ofNullable(Coordinates.builder().latitude(lat).longitude(lng).build());
   }
 
   private int driveMins(String serviceType) {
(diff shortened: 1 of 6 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -169,13 +169,13 @@ private static Address toAddress(Optional<AddressInfo> eeAddress) {
...
-  private static Coordinates toCoordinates(String patientIcn, GeocodingInfo geocodingInfo) {
+  private static Optional<Coordinates> toCoordinates(GeocodingInfo geocodingInfo) {
     BigDecimal lat = geocodingInfo.getAddressLatitude();
     BigDecimal lng = geocodingInfo.getAddressLongitude();
     if (lat == null || lng == null) {
-      throw new Exceptions.MissingGeocodingInfoException(patientIcn);
+      return Optional.empty();
     }
-    return Coordinates.builder().latitude(lat).longitude(lng).build();
+    return Optional.ofNullable(Coordinates.builder().latitude(lat).longitude(lng).build());
```
jblefkove-va: Don't need `ofNullable` here since this will never be null.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -259,11 +259,15 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
 
     Optional<GeocodingInfo> geocoding = geocodingInfo(eeResponse);
     if (geocoding.isEmpty()) {
-      throw new Exceptions.MissingGeocodingInfoException(request.patientIcn());
+      return response.build();
```
jblefkove-va: Let's add a log statement here (basically the same message that was previously on MissingGeocodingInfoException) so that there is some clue why we returned a partial response.

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/ApplicationTest.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class ApplicationTest {
+  @Test
+  public void contextLoads() {
+    /* Verifies that the application starts. */
+    assertTrue(true);
```
jblefkove-va: Is the `assertTrue` necessary, or would a blank test body also work?

jblefkove-va:
As I review this, it occurs to me that we should probably make the same change `OutdatedGeocodingInfoException` as well.

community-care-eligibility/src/test/java/gov/va/api/health/communitycareeligibility/service/ApplicationTest.java
jhulbert-va: I believe (and I can test this out if youd like) that one of our plugins flags empty methods as an issue.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -213,9 +210,9 @@ public CommunityCareEligibilityResponse search(
       // Strip newlines for Spotbugs
       log.info(
           "sessionId={}, patient={}, serviceType={}",
-          stripNewlines(optSessionIdHeader),
-          stripNewlines(patientIcn),
-          stripNewlines(serviceType));
+          LogSanitizer.sanitize(optSessionIdHeader),
+          LogSanitizer.sanitize(patientIcn),
+          LogSanitizer.sanitize(serviceType));
```
jblefkove-va: Let's statically import this. Also, you don't need to sanitize the other log statements below; these three here are the only ones SpotBugs was complaining about.

PR #94, opened 2020-06-05
API-1074 Add v0 to health-check

jblefkove-va:
https://vajira.max.gov/browse/API-987


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HealthController.java
```diff
@@ -85,7 +85,7 @@ private Health facilitiesHealth(@NonNull Instant time) {
    * @see #clearCacheScheduled()
    */
   @Cacheable("health")
-  @GetMapping(value = "/health")
+  @GetMapping(value = "/v0/health")
```
bschofield-va, line 88: why not support both?

jblefkove-va: Good point

PR #95, opened 2020-06-18
API-1273 JUnit 5

jblefkove-va:
https://vajira.max.gov/browse/API-1273


jblefkove-va:
Is that only for simple equality checks, or are we looking to drop AssertJ entirely?

jhulbert-va:
The only goal I'm aware of is to move to JUnit 5. I was just thinking for uniformity. I can be convinced to stick with AssertJ though if we think thats the better path forward.

jblefkove-va:
I'm inclined to stick with AssertJ. Even for simpler assertions, `assertThat(actual).isEqualTo(expected)` is more readable than `assertEquals(expected, actual)`

PR #105, opened 2020-07-23
CCE: Partial success status codes and documentation

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1312
Api 1312 partial success status codes and documentation updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,13 @@ and scheduled their appointment.
 The response includes the overall eligibility decision and the individual details
 (patient address, eligibility codes, nearby facilities, etc.) that were used to compute it.
 
+It should be noted that partial success states can occur. Partial success states can occur due to:
+```
+Address geocoding is not available
+Address geocoding is out of date compared to the users address
+Address geocoding is incomplete (e.g. missing latitude)
+```
+
 Sample request:
 
 ```
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Sample response:
       "website" : "https://www.va.gov"
     }
   ],
-  "eligible" : false
+  "eligible" : false,
+  "processingStatus" : successful
 }
 ```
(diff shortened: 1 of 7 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -264,6 +264,7 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
           .eligible(!codeStrings.contains("X"))
           .grandfathered(codeStrings.contains("G"))
           .noFullServiceVaMedicalFacility(codeStrings.contains("N"))
+          .processingStatus(CommunityCareEligibilityResponse.ProcessingStatus.successful)
```
jblefkove-va, line 267: Would it make sense to set this up on lines 254-259? Then you only need to override it if its not successful.

README.md
```diff
@@ -67,6 +67,13 @@ and scheduled their appointment.
 The response includes the overall eligibility decision and the individual details
 (patient address, eligibility codes, nearby facilities, etc.) that were used to compute it.
 
+It should be noted that partial success states can occur. Partial success states can occur due to:
+```
+Address geocoding is not available
+Address geocoding is out of date compared to the users address
+Address geocoding is incomplete (e.g. missing latitude)
```
jblefkove-va, line 74: Update the sample response to include the new field

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
taylorrassmannlibertyits: Yep! I'll make the change

bschofield-va:
I did not really review the code, other than the CODEOWNERS so you could move forward at your leisure.

PR #114, opened 2020-11-23
API-3753 Update swagger schema documentation to match current implementation

jnguy-va:
https://vajira.max.gov/browse/API-3753

Updated the swagger to match the README+Dev response as well as formatted the README to the current spacings.


taylorrassmannlibertyits (approved):

jnguy-va:
Before and after

image

darduda:
I took a look and this seems fine. My only question was I don't see mobile and active on all facilities returned. This might not be something to fix per se, but should it appear on all or only some?

PR #117, opened 2021-02-25
CCE now always returns ineligible when service type request is Primary Care

as-va:
https://vajira.max.gov/browse/API-5367

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
@@ -263,6 +263,10 @@ public class CommunityCareEligibilityV0ApiController {
             .noFullServiceVaMedicalFacility(false)
             .processingStatus(CommunityCareEligibilityResponse.ProcessingStatus.successful);
 
+    if (request.serviceType().equals(SERVICES_MAP.get("PrimaryCare"))) {
+      return response.eligible(false).build();
+    }
+
     List<String> codeStrings =
         eligibilityCodes.stream().map(c -> c.code()).collect(Collectors.toList());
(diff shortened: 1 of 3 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -263,6 +263,10 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
             .noFullServiceVaMedicalFacility(false)
             .processingStatus(CommunityCareEligibilityResponse.ProcessingStatus.successful);
 
+    if (request.serviceType().equals(SERVICES_MAP.get("PrimaryCare"))) {
```
taylorrassmannlibertyits, line 266: Around line 199 in this file, there is an `equalsIgnoreCase` when checking service type with primary care. maybe this should be used?

taylorrassmannlibertyits (approved):

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
as-va: The referenced line is comparing a string literal to a string argument and therefore needs `equalsIgnoreCase` to do a case-insensitive comparison.

This code is also comparing (indirectly) with a string literal, but the literal is used as a key into a case-insensitive map to retrieve the API internal representation of the service string. Basically routing through the `SERVICES_MAP` effectively ignores the case of the input parameter. This is the same way the `request.serviceType` field is initialized (lines ~222/235).

taylorrassmannlibertyits: Cool cool cool.

PR #118, opened 2021-02-26
CCE: Create mock PCMM interface

taylorrassmannlibertyits:
Set up mock endpoints for PCMM data with three different example returns.
https://vajira.max.gov/browse/API-5306

### DIFF
--- a/community-care-eligibility-mock-services/src/main/java/gov/va/api/health/communitycareeligibilitymockservices/MockServices.java
+++ b/community-care-eligibility-mock-services/src/main/java/gov/va/api/health/communitycareeligibilitymockservices/MockServices.java
@@ -28,6 +28,8 @@ import org.springframework.stereotype.Component;
 @Slf4j
 @Component
 public class MockServices {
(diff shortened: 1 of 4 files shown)


as-va (commented):
Looks good. One quick question before approval: do we know if `Active` and `Pending` (and no data) are the only assignment status responses we can get? I checked the ICD but it does not specify the values for this field other than to say it is a string.

taylorrassmannlibertyits:
> Looks good. One quick question before approval: do we know if `Active` and `Pending` (and no data) are the only assignment status responses we can get? I checked the ICD but it does not specify the values for this field other than to say it is a string.

There are two other results `Inactive` and `Cancelled` as well.

jnguy-va (approved):
New mock endpoints returning as expected.
Old CCE calls unchanged.

as-va (approved):
Confirmed with PCMM that `Inactive` and `Cancelled` states are internal states only. The only states the API will return are `Active`, `Pending`, and `No Data` which are covered by this PR.

PR #120, opened 2021-03-02
CCE: Move existing /nearby processing to async task

taylorrassmannlibertyits:
Moved nearby and other little bits of logic to nearby CompletableFuture thread.

https://vajira.max.gov/browse/API-5308


as-va:
Since you are modifying the response object inside the thread, we should ensure this is thread safe. Add these lines to the annotations of `CommunityCareEligibilityResponse`:
```
@Setter(onMethod=@__(@Synchronized))
@Getter(onMethod=@__(@Synchronized))
```

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -202,8 +202,84 @@ private int driveMins(String serviceType) {
...
-  private String requestNearbyFacilityResults() {
-    return "Nearby Facilities Results Stub";
+  private CommunityCareEligibilityResponse requestNearbyFacilityResults(
+      PatientRequest request,
+      CommunityCareEligibilityResponse.CommunityCareEligibilityResponseBuilder response,
+      GetEESummaryResponse eeResponse,
+      List<String> codeStrings) {
+
+    if (CollectionUtils.containsAny(codeStrings, asList("G", "N", "H"))) {
+      return response
```
as-va, line 212: This doesn't set eligibility, but it should be `True` here.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -202,8 +202,84 @@ private int driveMins(String serviceType) {
         : maxDriveMinsSpecialty;
   }
 
-  private String requestNearbyFacilityResults() {
-    return "Nearby Facilities Results Stub";
+  private CommunityCareEligibilityResponse requestNearbyFacilityResults(
```
as-va, line 205: We do not want or need to return a CCEResponse object here - for two reasons:
1. It is unnecessary - we pass in the response object from the point of call, so the operations here take place on that object which is already in the scope of the caller so we don't need to pass another reference to it back out again.
2. We need to use the return type to handle the eligibility flag - Since the design has shifted to a tri-state eligibility flag (true/false/null), we cannot simply set true before threads and overwrite within as there are now two possible other values. So we need to return the eligibility determination from each thread and have the composer take the returns to set the final eligibility flag in the response.

Side effect: need to take out all response.build() lines inside this function. Do a single build once both threads have completed.

as-va (changes requested):

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
taylorrassmannlibertyits: ok, at first you said to not set global eligibility to true in the jira issue. I will make the change.

as-va: Right, that was more just a heads up since we are changing how the eligibility reporting is done in my other comment so that this return spot didn't fall through the cracks.

as-va (approved):

PR #122, opened 2021-03-12
Api 5310 merge pcmm and nearby task results

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-5310

### DIFF
--- a/community-care-eligibility-tests/src/test/java/gov/va/api/health/communitycareeligibility/tests/AdvancedSearchIT.java
+++ b/community-care-eligibility-tests/src/test/java/gov/va/api/health/communitycareeligibility/tests/AdvancedSearchIT.java
@@ -38,7 +38,7 @@ public class AdvancedSearchIT {
             "v0/eligibility/search?patient=%s&serviceType=%s", "1012845331V153043", "Cardiology");
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
-    //    assertThat(response.eligible()).isTrue();
+    assertThat(response.eligible()).isTrue();
   }
 
   @Test
@@ -49,7 +49,7 @@ public class AdvancedSearchIT {
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
     assertThat(response.nearbyFacilities()).isEmpty();
-    //    assertThat(response.eligible()).isTrue();
+    assertThat(response.eligible()).isTrue();
   }
 
   @Test
@@ -60,7 +60,7 @@ public class AdvancedSearchIT {
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
     assertThat(response.noFullServiceVaMedicalFacility()).isTrue();
-    //    assertThat(response.eligible()).isTrue();
+    assertThat(response.eligible()).isTrue();
   }
 
   @Test
@@ -71,7 +71,7 @@ public class AdvancedSearchIT {
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
     assertThat(response.patientAddress()).isNull();
-    //    assertThat(response.eligible()).isNull();
+    assertThat(response.eligible()).isNull();
   }
 
   @Test
@@ -82,7 +82,7 @@ public class AdvancedSearchIT {
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
     assertThat(response.patientCoordinates()).isNull();
-    //    assertThat(response.eligible()).isNull();
+    assertThat(response.eligible()).isNull();
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -419,24 +419,28 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
...
-              //              } else {
-              //                response.eligible(pcmmRequestResult && nearbyRequestResult);
-              //              }
-              return "Stub: " + pcmmRequestResult + ":" + nearbyRequestResult;
+
+              // Priority order:
+              // If any result is false return false
+              // Else if any result is null return null
+              // else return true (since this must be TRUE at this point)
+              if ((pcmmRequestResult != null && !pcmmRequestResult)
```
as-va, line 431: There is nothing wrong with this code and it accomplished the desired behavior correctly. That said, it is unfortunate this has to be four checks including six terms (if you count the nulls) in multiple sub groups to do a simple boolean comparison (feature of the language).

I think I would have a slight preference for 

    if (Boolean.FALSE.equals(pcmmRequestResult) || 
        Boolean.FALSE.equals(nearbyRequestResult)) {

as this cuts the complexity of the statement back down to the actual intent of the check without the housekeeping of the embedded null checks or the nesting of parenthesis to accomplish the task.

Not requesting a change here, just pointing out a different stylistic option.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
```diff
@@ -419,24 +419,28 @@ private CommunityCareEligibilityResponse search(PatientRequest request) {
...
+              }
+              return null;
             });
 
-    // Stub before actual results processing
-    System.out.println(
-        "Results of PCMM and Nearby Facilities calls: "
-            + combinedPcmmAndNearbyResultsFuture.get(10, TimeUnit.SECONDS));
+    // If this takes longer than 10 seconds, it will probably not complete at all
+    combinedPcmmAndNearbyResultsFuture.get(10, TimeUnit.SECONDS);
```
as-va, line 443: I don't think this works this simply. `get` raises multiple checked exception types that you aren't handling because the `SneakyThrows` is masking them, but if you actually trigger this timeout, the future doesn't just return, it throws a `TimeoutException` which isn't being caught.

as-va, line 443: You could add a handler here for the `get` timeout, but instead I think I'd pass responsibility back up the the original Futures and add a `completeOnTimeout` registration where they are defined. That can set up a timeout on each thread _and_ specify a default value to return in case the timeout is triggered so that the combining Future can still function as designed with no changes.

as-va (changes requested):

community-care-eligibility-tests/src/test/java/gov/va/api/health/communitycareeligibility/tests/AdvancedSearchIT.java
```diff
@@ -38,7 +38,7 @@ void patientEligibleDueToHardship() {
             "v0/eligibility/search?patient=%s&serviceType=%s", "1012845331V153043", "Cardiology");
     CommunityCareEligibilityResponse response =
         makeRequest(request, 200).expectValid(CommunityCareEligibilityResponse.class);
-    //    assertThat(response.eligible()).isTrue();
+    assertThat(response.eligible()).isTrue();
```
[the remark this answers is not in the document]
taylorrassmannlibertyits: One of the main reasons why I didn't want to make those changes for the previous PR. Will update in the next comment.

community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/CommunityCareEligibilityV0ApiController.java
taylorrassmannlibertyits: Even if the exception isn't caught, isn't that a good thing? Don't we want to fail on a timeout?

taylorrassmannlibertyits: If either thread has a timeout, shouldn't we fail the call? I disagree here. I think the timeout should be in the combined call.

as-va, line 443: There should be a new test to cover the expected behavior when a timeout occurs.

as-va: In the most aggressive approach you could say we should return a `4XX` error to the caller (with reason message), but this would be a fatal condition for the request handler which I believe would return a `5XX` error response which seems too extreme a response.

as-va: I can see a logic for that argument, but we've already been instructed that if we do not get a response from PCMM to handle that by setting PCMM eligibility to `null`. It seems to me that the PCMM call not returning is a sub form of the 'failure to get PCMM results' handler and should be handled the same way.

as-va (approved):

PR #123, opened 2021-03-12
Added PCMM health monitoring

as-va:
https://vajira.max.gov/browse/API-5489

### DIFF
--- a/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HealthController.java
+++ b/community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HealthController.java
@@ -28,14 +28,18 @@ public class HealthController {
 
   private final FacilitiesClient facilitiesClient;
 
+  private final PcmmClient pcmmClient;
+
   private final AtomicBoolean hasCachedRecently = new AtomicBoolean(false);
 
   @Builder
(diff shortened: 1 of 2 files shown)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/HealthController.java
```diff
@@ -59,20 +63,23 @@ public void clearCacheScheduled() {
   }
 
   private Health eeHealth(@NonNull Instant time) {
+    HttpStatus status = HttpStatus.OK;
```
taylorrassmannlibertyits, line 66: Why change this from multiple returns to one at the end of this method?

taylorrassmannlibertyits: For consistency, I recommend changing the other health checks, or changing this one back to the original format.

as-va: I just tried to simplify it a bit since the function is only a few lines long and with the addition of extra catch blocks _half_ of them were returns. I can update the others as well.

taylorrassmannlibertyits (approved):

PR #128, opened 2021-07-08
Adding ExampleObjects to the Parameters

xenkidu-va:
### [API-8480](https://vajira.max.gov/browse/API-8480)
We are adding ExampleObjects to the Parameters.

### DIFF
--- a/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
+++ b/community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
@@ -190,7 +190,7 @@ public interface CommunityCareEligibilityService {
                   "Optional extended drive-radius to include more VA medical facilities in response"
                       + " (Does not change overall eligibility."
                       + " Must exceed standard drive time for service-type.)",
-              examples = @ExampleObject(name = "extendedDriveMin", value = "45"))
+              examples = @ExampleObject(name = "extendedDriveMin", value = "40"))
           @Max(90)
           Integer driveMin);
 }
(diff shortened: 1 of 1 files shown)


community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
```diff
@@ -186,7 +189,8 @@ CommunityCareEligibilityResponse search(
               description =
                   "Optional extended drive-radius to include more VA medical facilities in response"
                       + " (Does not change overall eligibility."
-                      + " Must exceed standard drive time for service-type.)")
+                      + " Must exceed standard drive time for service-type.)",
+              examples = @ExampleObject(name = "extendedDriveMin", value = "45"))
```
as-va, line 193: Only multiples of 10 between 10 and 90 are allowable values. This example value will not work and will break the OAS testing.

xenkidu-va: Good eye! I went ahead and updated.

taylorrassmannlibertyits (approved):

community-care-eligibility-api/src/main/java/gov/va/api/health/communitycareeligibility/api/CommunityCareEligibilityService.java
as-va: Oops, one more note on this - the extended drive time should be greater than the default drive time. Default for primary care is 30 mins, specialty care is 60 mins. This example is using `Gastroenterology` for the service type, which is specialty care, so the extended drive time should be greater than 60.

xenkidu-va: Sounds good, updated to 70.

as-va: Thanks, I'm done griping now!

as-va (approved):

PR #132, opened 2021-07-19
API-8622 Fixed patient example value to be an ICN

as-va:
https://vajira.max.gov/browse/API-8622


taylorrassmannlibertyits:
Why the removal of `@ExampleObject` and change to `example` ?

as-va:
> 
> 
> Why the removal of `@ExampleObject` and change to `example` ?

Since those two parameters are simple types and are both required, `examples` is overkill and `example` is more appropriate.
`ExampleObject` was removed because it is the type object for `examples`.

taylorrassmannlibertyits (approved):

PR #134, opened 2021-07-21
add pcmm properties to make-configs.sh

jblefkove-va:
Update `make-configs.sh` to set the pcmm properties. Also make facilities secret properties optional, using `community-care-eligibility-mock-services` values as defaults.


src/scripts/make-configs.sh
```diff
@@ -43,12 +43,11 @@ do
 done
 
 echo "Loading secrets: $SECRETS"
-[ ! -f "$SECRETS" ] && usage "File not found: $SECRETS"
```
as-va, line 46: I see removing the restriction on the having `secrets`, but shouldn't the if check on file existence remain before accessing the file on the next line?

jblefkove-va: Good point

taylorrassmannlibertyits (approved):

PR #145, opened 2021-11-05
API-9375 Adding additional ICNS to readme for differing PACT Statuses…

JamesDanielBell-VA:
https://vajira.max.gov/browse/API-9375

- Adding ICN's to the CCE README with following PACT Statuses:
1. Active
2. Pending
3. None


README.md
```diff
@@ -50,6 +50,14 @@ The medical service type is one of:
 The response includes the overall eligibility decision and the individual details
 (patient address, eligibility codes, nearby facilities, PACT status, etc.) that were used to compute it.
 
+Sample ICNS with differing PACT Statuses:
```
as-va, line 53: Let's make the `S` lowercase so it does not look like part of the acronym.
`ICNs`

as-va (approved):

PR #146, opened 2021-12-02
API-11610 CCE: Add time out logic into backend calls

giangnguyen2:
Added timeout to rest template and jax ws request per [API-11610](https://vajira.max.gov/browse/API-11610)


community-care-eligibility/src/main/java/gov/va/api/health/communitycareeligibility/service/InsecureRestTemplateProvider.java
```diff
@@ -43,20 +31,21 @@ private CloseableHttpClient httpClientWithoutSsl() {
...
-    if (readTimeout != null) {
-      log.info("Setting read timeout to {} seconds.", readTimeout);
-      restTemplateBuilder = restTemplateBuilder.setReadTimeout(readTimeout);
+
+    if (readTimeout > 0) {
+      log.info("Setting read timeout to {} milliseconds.", readTimeout);
     }
-    return restTemplateBuilder
-        .requestFactory(bufferingRequestFactory(httpClientWithoutSsl()))
-        .build();
```
giangnguyen2, line 60: I made changes for this bean because the previous code doesn't set timeout even when values are supplied in application properties for some reasons.

as-va (approved):

PR #159, opened 2022-12-06
Updated entrypoint to support dynamic api path & port

ChrisBratti-VA:


community-care-eligibility-tests/src/main/docker/entrypoint.sh
```diff
@@ -124,7 +124,13 @@ setupForAutomation() {
...
   addSystemProperty sentinel "$SENTINEL_ENV"
-  addSystemProperty sentinel.community-care-eligibility.api-path "$COMMUNITY_CARE_ELIGIBILITY_API_PATH"
+  if [[ "$DEPLOYMENT_D2" == "true" ]];
+  then
+    [ -n "${COMMUNITY_CARE_API_PATH:-}" ] && addSystemProperty sentinel.community-care-eligibility.api-path "${COMMUNITY_CARE_API_PATH}"
+    [ -n "${COMMUNITY_CARE_TEST_PORT:-}" ] && addSystemProperty sentinel.community-care-eligibility.port "${COMMUNITY_CARE_TEST_PORT}"
+  else
+    addSystemProperty sentinel.community-care-eligibility.api-path "$COMMUNITY_CARE_ELIGIBILITY_API_PATH"
+  fi
   addSystemProperty sentinel.community-care-eligibility.url "$COMMUNITY_CARE_ELIGIBILITY_URL"
```
jhulbert-va, line 134: This url should change based on d1 and d2.

D1: `https://$K8S_LOAD_BALANCER`
D2: `$DEPLOYMENT_TEST_PROTOCOL://$DEPLOYMENT_TEST_HOST`

community-care-eligibility-tests/src/main/docker/entrypoint.sh
```diff
@@ -124,7 +124,13 @@ setupForAutomation() {
 
   addSystemProperty access-token "$TOKEN"
   addSystemProperty sentinel "$SENTINEL_ENV"
-  addSystemProperty sentinel.community-care-eligibility.api-path "$COMMUNITY_CARE_ELIGIBILITY_API_PATH"
+  if [[ "$DEPLOYMENT_D2" == "true" ]];
+  then
+    [ -n "${COMMUNITY_CARE_API_PATH:-}" ] && addSystemProperty sentinel.community-care-eligibility.api-path "${COMMUNITY_CARE_API_PATH}"
+    [ -n "${COMMUNITY_CARE_TEST_PORT:-}" ] && addSystemProperty sentinel.community-care-eligibility.port "${COMMUNITY_CARE_TEST_PORT}"
```
jhulbert-va, line 130: Port is handed to you by d2. It shouldn't be respecified here as a CCE specific var: `${DEPLOYMENT_TEST_PORT}`

community-care-eligibility-tests/src/main/docker/entrypoint.sh
```diff
@@ -124,7 +124,13 @@ setupForAutomation() {
 
   addSystemProperty access-token "$TOKEN"
   addSystemProperty sentinel "$SENTINEL_ENV"
-  addSystemProperty sentinel.community-care-eligibility.api-path "$COMMUNITY_CARE_ELIGIBILITY_API_PATH"
+  if [[ "$DEPLOYMENT_D2" == "true" ]];
+  then
+    [ -n "${COMMUNITY_CARE_API_PATH:-}" ] && addSystemProperty sentinel.community-care-eligibility.api-path "${COMMUNITY_CARE_API_PATH}"
```
jhulbert-va, line 129: API Path will not be specific to d1 vs d2 deployments. It can be outside this if statement and should be specified in the testvars file.

community-care-eligibility-tests/src/main/docker/entrypoint.sh
ChrisBratti-VA: Ahh okay that makes sense, I saw that being done in Data Query but assumed those were Data Query specific variables, I'll update

community-care-eligibility-tests/src/main/docker/entrypoint.sh
ChrisBratti-VA: I actually had that earlier but removed it because I wasn't sure if it needed to be done for CCE, I'll go ahead and put that back

community-care-eligibility-tests/src/main/docker/entrypoint.sh
ChrisBratti-VA: Would we not want the D1 & D2 deployments to have different API paths? K8s is set up to reach out to `/community-care/` and ECS is set up to reach out to `/community-care-ecs/`. Would we want the K8s build's tests also hitting `/community-care-ecs/`?

ChrisBratti-VA: Whoops never mind I get what you're saying now, since that var is only being set in the D2 build, D1 will use the default which is currently set to `/community-care/`. I'll go ahead and take it out of the if statement

jhulbert-va: Yes, that or you could tell the k8s testvars to change the path as well. By leaving it out of the if, you have options. I would probably put it in an if that checks the vars existence before it adds it as a property though.

community-care-eligibility-tests/src/main/docker/entrypoint.sh
```diff
@@ -124,8 +124,17 @@ setupForAutomation() {
...
   addSystemProperty access-token "$TOKEN"
   addSystemProperty sentinel "$SENTINEL_ENV"
-  addSystemProperty sentinel.community-care-eligibility.api-path "$COMMUNITY_CARE_ELIGIBILITY_API_PATH"
-  addSystemProperty sentinel.community-care-eligibility.url "$COMMUNITY_CARE_ELIGIBILITY_URL"
+  if [[ "$DEPLOYMENT_D2" == "true" ]];
+  then
+    CCE_URL="$DEPLOYMENT_TEST_PROTOCOL://$DEPLOYMENT_TEST_HOST"
+  else
+    CCE_URL="https://$K8S_LOAD_BALANCER"
+    DEPLOYMENT_TEST_PORT="443"
```
jhulbert-va, line 132: This is already the default https port. I don't think theres a need to set it here, but I'm not gonna hold up the PR over it.

as-va (approved):

PR #167, opened 2023-02-14
API-19523-Resolving-WASA-exception-discrepancies

JamesDanielBell-VA:
https://vajira.max.gov/browse/API-19523

- Adding logic to deal with MethodArgumentTypeMismatchException


community-care-eligibility/src/main/resources/application.properties
```diff
@@ -22,15 +22,15 @@ jax-ws.connection-timeout=5000
 jax-ws.read-timeout=5000
 
 va-facilities.api-key=unset
-va-facilities.url=unset
+va-facilities.url=http://localhost:8319
```
as-va, line 25: Looks like these changes were for your dev profile but shouldn't be committed to the master properties file. `make-configs` should populate these unset fields into a dev profile for your local use.

JamesDanielBell-VA: Thank you kindly. Pushed out the correction

as-va (approved):

PR #171, opened 2023-05-11
Java 17 upgrade

ChrisBratti-VA:


as-va (changes requested):
Let's change the `parent` reference version to a range to make maintenance easier going forward. Instead of explicitly setting to `13.0.5` (which is already out of date before this PR merges) lets have the pom accept anything in the 13.0.x line automatically. Then when the parent bumps to 13.1, automatic updates will stop and we can come back manually and ensure the changes are compatible with the API before blindly updating.

The version can be set to a range with this notation: `<version>[13.0.5-13.1.0)</version>`. The bracket says include this version in the range and the parenthesis says use anything up to but not including this version itself. Maven will then resolve this range on build to find the highest numbered version released within that range and use it in the build.

as-va (approved):

